PR #9, opened 2017-08-31
substitutions: put substitutions in separate files

tobsan:
This way, both the blueprint project, and any deployment of it, can add
its own substitutions without having to fiddle around in the
configuration files, plus the deployment can use the substitutions from
the blueprint.


GordanM (approved):

thenor (approved):

rpannek (changes requested):
In this version it is not clear to me how I'm supposed to use it.

tobsan:
@rpannek Updated with usage example

breakreturn (approved):

rpannek (approved):

PR #10, opened 2017-09-06
restructure with data taken from PELUX

tobsan:
Note: Once this gets merged, the idea is to do a PR for the PELUX Software Factory to make use of this new structure.

Also describes the idea of the blueprint and that it should be used with
a deployment so that it is clear why it may not seem useful.


tobsan:
Structure of articles is not final but will probably change, but it is a start at least.

rpannek (approved):

docs/articles/baseplatform/index.rst
```diff
@@ -0,0 +1,19 @@
...
+         articles
+.. toctree::
+    prerequisites.rst
+    using-the-repo-tool.rst
+    repo-manifests.rst
+    writing-image.rst
+    creating-sdk.rst
+    creating-a-new-layer.rst
+
+.. .. include:: prerequisites.rst
```
breakreturn: Why are these left in here?

docs/articles/intro/index.rst
```diff
@@ -0,0 +1,66 @@
...
+part of the project. The blueprint can then be used as a base and then be extended with
+project-specific info, or one can just pick and choose the relevant parts from the blueprint.
+
+.. only:: blueprint
+
+    If you are reading this, it means you are reading the blueprint and not a deployment, meaning
+    that some links or names may look weird. Don't be alarmed, this is (mostly) by design.
+
+A nice way of using this blueprint in a project-specific setting is to use it as a *git submodule*
+from whichever git contains the project-specific Software Factory. Our reference deployment *PELUX*
```
breakreturn: The reference Deployment is "PELUX Baseline SWF Deployment"

docs/articles/intro/index.rst
```diff
@@ -0,0 +1,66 @@
...
+How to use this blueprint
+=========================
+The blueprint is not intended to be particularly useful on its own, rather it is a collection of
+generic descriptions, processes, how-tos et cetera.
+
+The typical user of this blueprint is someone who wants to run a project where a Software Factory is
+part of the project. The blueprint can then be used as a base and then be extended with
+project-specific info, or one can just pick and choose the relevant parts from the blueprint.
+
+.. only:: blueprint
```
breakreturn: I don't like this. Could we write the content in general as  "the blueprint" and not "this blueprint" and assume that the blueprint is something that co-exists with a Deployment at most times. The few that reads the Blueprint in isolation should have gotten the description of the structure as part of the Blueprint anyway and will know that there is a Deployment missing.

The section that describes the structure to the reader, including nomenclature etc, doesn't need to be part of this change, but this change could assume there will be that kind of info later.

docs/articles/intro/index.rst
```diff
@@ -0,0 +1,66 @@
...
+
+    If you are reading this, it means you are reading the blueprint and not a deployment, meaning
+    that some links or names may look weird. Don't be alarmed, this is (mostly) by design.
+
+A nice way of using this blueprint in a project-specific setting is to use it as a *git submodule*
```
breakreturn: In general I think we should write how to do this, not recommend one way of doing it. The difference is often small, but the docs need to read consistently throughout, and this is the manual of how to actually do things. When a rationale of why we don something in a specific way or when it's relevant to compare different approaches that could be separated from the other text.

breakreturn (changes requested):
Nice. Have some comments on the generic "this is the Blueprint" texts.

docs/articles/baseplatform/index.rst
tobsan: Oops, my bad!

tobsan:
@breakreturn updated after your comments.

docs/articles/intro/index.rst
```diff
@@ -0,0 +1,61 @@
...
+The blueprint is not intended to be particularly useful on its own, rather it is a collection of
+generic descriptions, processes, how-tos et cetera.
+
+The typical user of the blueprint is someone who wants to run a project where a Software Factory is
+part of the project. The blueprint can then be used as a base and then be extended with
+project-specific info, or one can just pick and choose the relevant parts from the blueprint.
+
+We recommend using the blueprint in a project-specific setting as a *git submodule* from whichever
+git contains the project-specific Software Factory. Our reference deployment *PELUX Baseline SWF
+Deployment* does exactly that.
```
breakreturn: Why not link now already? We have the Baseline Deployment available publicly.

PR #18, opened 2017-10-03
Rename all files with toctrees to .trst

rpannek:
To be able to use files from the blueprint in other documentation projects, those files can not be used in any toctree in the blueprint. This is because Sphinx parses all the files in all directories and looks for toctrees. If it finds one it tries to create the sidebar from them. If one file is found in two seperate toctrees at the same time Sphinx gets confused and crashes leaving a blank sidebar.

In the conf.py file we can set up which suffixes we want Sphinx to traverse to find all files. This patch atts the suffix .trst to the list and renames all the files which have a toctree to .trst too. This way the blueprint can be build on itself with the .trst files, but the real project doesn't pick them up because it only has .rst in their list.

Another patch sets the proj_name substitution to semething sensible
    but this will be most probably be overwritten higher up in the real
    project.


tobsan (changes requested):
Looks good. Can we mention somewhere in a README or something to make sure all toctree files are named `.trst` in the future?

rpannek:
Ok, I added a paragraph about that.

jonte (approved):
Nice!

PR #19, opened 2017-10-04
variant-concept: Add article

jonte:


GordanM (approved):

breakreturn:
I think the content is good. What I think we should do is to clarify throughout that this is the Yocto domain only. The heading and file name and the way the text reads etc.

The reason is that I believe the concept of variant is larger for the projects using this than "just" the Yocto  aspects and we should take height for that already now in the nomenclature and wordings.

breakreturn (changes requested):
See comment in Conversations...

rpannek (approved):

breakreturn (approved):

docs/articles/baseplatform/variant-concept-in-yocto.rst
```diff
@@ -0,0 +1,49 @@
...
+[#templateconf]_, and this scheme is also used in the Software Factory.
+
+.. _recommended-variant-management:
+
+Recommended variant management
+------------------------------
+
+The recommended way to handle variants is to keep one directory per variant in
+your layer. The variant directories should go into
+``<meta-yourlayer>/conf/<variant name>``. Each variant should supply a
```
tobsan, line 24: Actually, the current use (in meta-pelux) is to have `meta-pelux/meta-<hardware>-extras/conf-<variant>`, and I guess at least the hardware specific stuff makes sense to separate like that because one might want to include additional recipes. For meta-pelux, can we compromise and use `meta-pelux-meta-<hardware>-extras/conf/<variant>/` ?

tobsan: With this, the variants would be software variants, nested under hardware variants.

tobsan (commented):
Just had one comment.

docs/articles/baseplatform/variant-concept-in-yocto.rst
jonte: For my current project, I have used a variant name with the format "hardwarevariant_softwarevariant" - an example would be "rpi3_nogfx" - this means I currently do not nest in the way you describe..

tobsan: Okay, but that would mean that any BSP specific recipes will be duplicated, right? If you have `rpi3_nogfx` and `rpi_qt` and both use some kernel bbappend?

jonte: I think the selection of variants (for instance, maybe the project will only ever use one hardware) should be left up to the project - some projects may only need software variants.

tobsan: While I agree with that, we set a norm by writing in a certain way, and that way should be the "standard" way. That's why I'm picky about how we formulate it here.

jonte: Well, recipes & bbappends are not necessarily grouped by variants, I suppose? For instance, a set of Qt-related bbappends would be used by all hardwares, and the recipe itself would contain MACHINE-specific sections.

My idea with these variants is mostly just for local.conf and bblayers.conf files.

jonte: BTW, my #2 comment in this thread was not a reply to your comment #3.. It didn't show up in my UI.

jonte: I guess my main point is that BSP != variant. You can have several variants using the same BSP.

PR #20, opened 2017-10-05
First draft of policy for public binaries

DunderRoffe:
This is ment as a start of discussion on the text. Please feel free to
push commits to this pull request. I will squash this pull request before
it is merged.

Signed-off-by: Viktor Sjölind <viktor.sjolind@pelagicore.com>

### DIFF
--- a/docs/articles/licensing/publishing-binaries.rst
+++ b/docs/articles/licensing/publishing-binaries.rst
@@ -1,21 +1,23 @@
-Policy for publishing build artifacts
-===================================== 
+Publishing binary images
+========================== 
 
-Since there are parts of the code which are licensed as GPL we are
-obligated to provide complete and corresponding source code together with
-publication of any binary artifact on request. Furthermore, the license
-demands that the source code is available for three years.
+We release binary images of our products that are licensed under the 
+[https://www.gnu.org/licenses/old-licenses/gpl-2.0.html](GNU Public License).
+This is considered "distribution" under the GPL. When distributing binaries, 
+or object code, covered under the GPL "complete and corresponding" source code 
+is required to be provided as well. We comply with these obligations 
+by publishing source code on GitHub and through our continuous integration 
+system (CI).
 
-Consequences
-------------
-In order to not violate the license, all public CI systems need to publish
-corresponding source code together with _any_ build artifact. This includes
-things like yocto sstatecache.
+In order to comply with all FOSS software licenses, public CI systems must 
+plublish complete and corresponding source code in conjunction with the publishing
+of object code. This include a yocto sstatecache, binary images for hardware 
+targets, and binary packages, like rpms or debs.
 
-Desition
+Decision
 --------
-When doing a release of a full systems or product this is not that a
-problem (assuming that there is no external propriatary code).
+When doing a release of a full system or product this is not that a
+problem (assuming that there is no external proprietary code).
 
 However, in a context where publishing is done often the requirement
 to store the source code leads to disk space issues. An example is when
@@ -25,4 +27,4 @@ to be available for each published iteration of each publish binary.
 Conclusion
 ----------
 To avoid unforseen cost of storing old source code, do not publish any
-binaries publically without first gotten the approval to do so.
+binaries publically without first receiving approval.
--- a/docs/articles/licensing/publishing-binaries.rst
+++ b/docs/articles/licensing/publishing-binaries.rst
@@ -1,0 +1,28 @@
+Policy for publishing build artifacts
+===================================== 
+
+Since there are parts of the code which are licensed as GPL we are
+obligated to provide complete and corresponding source code together with
+publication of any binary artifact on request. Furthermore, the license
+demands that the source code is available for three years.
+
+Consequences
+------------
+In order to not violate the license, all public CI systems need to publish
+corresponding source code together with _any_ build artifact. This includes
+things like yocto sstatecache.
+
+Desition
+--------
+When doing a release of a full systems or product this is not that a
+problem (assuming that there is no external propriatary code).
+
+However, in a context where publishing is done often the requirement
+to store the source code leads to disk space issues. An example is when
+doing rapid prototyping of applications to a system where the code needs
+to be available for each published iteration of each publish binary.
+
+Conclusion
+----------
+To avoid unforseen cost of storing old source code, do not publish any
+binaries publically without first gotten the approval to do so.


jeremiah:
I commited directly to the branch since I edited the page directly in GitHub. Do we want to allow this? Or should I follow the same recommendations we use for commiting code?

breakreturn:
@jeremiah Like code. You should comment on the PR to request changes and the author should choose how to incorporate them so the history can be managed by the author. Any changes from you should come as PRs from your fork. 

I suggest working on some document together in cases where there is a lot of changes expected, e.g. when working jointly on a body of text rather than code. For example on the wiki. Should make it easier to do quick iterations on the text IMO and make it easier to maintain a fairly nice history.

docs/articles/licensing/publishing-binaries.rst
```diff
@@ -0,0 +1,30 @@
+Publishing binary images
+========================== 
+
+We release binary images of our products that are licensed under the
```
breakreturn, line 4: In the Blueprint context we are not releasing anything.  In this part of the SWF we should write about the effects certain conditions and actions have in a general sense. The different Deployments (e.g. PELUX Baseline) should go into details about the concrete effects licenses have on the project, e.g. how it affects the releases or CI&CD jobs.

docs/articles/licensing/publishing-binaries.rst
```diff
@@ -0,0 +1,30 @@
...
+by publishing source code on GitHub and through our continuous integration 
+system (CI).
+
+In order to comply with all FOSS software licenses, public CI systems must 
+plublish complete and corresponding source code in conjunction with the publishing
+of object code. This include a yocto sstatecache, binary images for hardware 
+targets, and binary packages, like rpms or debs.
+
+Decision
+--------
```
breakreturn, line 18: I don't think we should have headings that are about decisions. The net outcome of whatever decisions that have been made during development should be documented instead. In some cases a rationale is relevant to provide context and guidance for projects though.

docs/articles/licensing/publishing-binaries.rst
```diff
@@ -0,0 +1,30 @@
...
+When doing a release of a full system or product this is not that a
```
breakreturn, line 19: This sentence is a bit off. I don't understand it.

docs/articles/licensing/publishing-binaries.rst
```diff
@@ -0,0 +1,30 @@
...
+problem (assuming that there is no external proprietary code).
+
+However, in a context where publishing is done often the requirement
+to store the source code leads to disk space issues. An example is when
+doing rapid prototyping of applications to a system where the code needs
+to be available for each published iteration of each publish binary.
+
+Conclusion
```
breakreturn, line 27: The content of the complete article should be the conclusion. Unless we're specifically writing a rationale or some in-depth analysis of a topic, then a conclusion might be separate from the content.

breakreturn (changes requested):

breakreturn:
@jeremiah Sorry, I see that @DunderRoffe asked for commits directly to this and intends to clean up later.

jeremiah:
I think @breakreturn is right. There is a lot of churn in editing text via PRs. Its noisy. I'd prefer a wiki page, so I set one up. Link shared internally.

docs/articles/licensing/publishing-binaries.rst
tobsan: On the other hand, this sentence can easily be adapted to describe what happens when we do something considered distribution by the GPL. "When we release...".

However, it would be nice to have this be more general description of how licenses affect what we need to do when doing binary releases.

docs/articles/licensing/publishing-binaries.rst
```diff
@@ -0,0 +1,30 @@
...
+We release binary images of our products that are licensed under the 
+[https://www.gnu.org/licenses/old-licenses/gpl-2.0.html](GNU Public License).
+This is considered "distribution" under the GPL. When distributing binaries, 
+or object code, covered under the GPL "complete and corresponding" source code 
+is required to be provided as well. We comply with these obligations 
+by publishing source code on GitHub and through our continuous integration 
+system (CI).
+
+In order to comply with all FOSS software licenses, public CI systems must 
+plublish complete and corresponding source code in conjunction with the publishing
```
tobsan, line 13: publish*

docs/articles/licensing/publishing-binaries.rst
```diff
@@ -0,0 +1,30 @@
...
+of object code. This include a yocto sstatecache, binary images for hardware
```
tobsan, line 14: Actually, there is a way in yocto to generate a `SOURCE_MIRROR`, which we should probably use when doing a release. With such a mirror, one could reproduce a build without having network access at all - meaning that it will contain complete and corresponding source code :-)

tobsan (changes requested):
In general I agree with @breakreturn here, but I added some comments of my own.

docs/articles/licensing/publishing-binaries.rst
breakreturn: The thing is that in some cases the source code needs to be potentially delivered to the user for years after the release, so the "problem" we have is that we don't want to trigger the clause at all too often.

tobsan: Of course, I just meant that instead of saving the sstate cache, we could create a source mirror, whenever the clause is triggered.

jeremiah: I prefer to use the yocto download dir but I guess a source mirror is good too. It might speed up builds actually.

tobsan: @jeremiah Are you sure using the download dir yields EXACTLY the same result though? I mean the very idea of a SOURCE_MIRROR is so that you can build without any network access (ignoring AUTOREV recipies, of course).

breakreturn: It should be possible to tie this together with the reproducibility use-cases in general which are part of releasing. If we can use the same archive to provide sources as we use to reproduce releases then I think we should align this with https://github.com/Pelagicore/software-factory-blueprint/pull/23

jeremiah: @tobsan Sorry for dropping this thread. Here is my understanding -- the SOURCE_MIRROR is for building without network access as you say, but the download dir (DL_DIR) is for making available the source code and the build toolchain. From Yocto's docs: http://www.yoctoproject.org/docs/2.4/mega-manual/mega-manual.html#providing-the-source-code

This is the approach I believe we should use since it appears to be the easiest from a compliance perspective.

breakreturn:
So what's up with this PR? Is anyone still nursing this along in a discussion somewhere?

DunderRoffe:
I do not think this is relevant anymore.

PR #22, opened 2017-10-06
describes how repo works

### LINKED ISSUE #21: Remove "Bla bla" text from blueprint
In my opinion if we don't have something meaningfull written about a topic then we should not create any placeholder pages but instead just not have anything about it. Otherwise in the table of contents it looks like we have info but when you click on it it just says "Bla bla". One example is:

https://github.com/Pelagicore/software-factory-blueprint/blob/master/docs/articles/baseplatform/repo-manifests.rst

This should be removed.

tobsan:
Closes #21 


docs/articles/baseplatform/repo-manifests.rst
```diff
@@ -1,6 +1,121 @@
...
+* Default settings for revision etc.
+* Separate remotes and actual repos.
+
+Usage in yocto
+--------------
+When doing yocto builds, the repo tool is useful to have the various yocto
+layers synced towards the same yocto release (typically, a poky release).
+
+Revisions in repo manifests can be branch names, tags, or commit hashes. For
+maximum reproducability, commit hashes is a good solution, but in order to be
```
rpannek: commit hashes is -> commit hashes are

docs/articles/baseplatform/repo-manifests.rst
```diff
@@ -1,6 +1,121 @@
...
+branches in the manifest repo, where the manifest in each branch would then
+point to the corresponding branch (or a commit on that branch) in the layers
+used.
+
+Example
+^^^^^^^
+Below is an example of a manifest from PELUX [#pelux-manifests]_ showcasing the
+format.
+
+.. code-block:: xml
```
rpannek: I think I'd prefare `.. literalinclude:: filename` instead. In theory one could use the pelux-manifests repo as a submodule and use the actuall file, but it might be too much hassle.

GordanM (approved):

rpannek (changes requested):

docs/articles/baseplatform/repo-manifests.rst
tobsan: Good point. But on that topic: do we want the FULL manifest? I was considering shortening it a bit actually for brevity.

rpannek: I wouldn't mind if it was shorter.

docs/articles/baseplatform/repo-manifests.rst
```diff
@@ -1,6 +1,121 @@
...
 =========
 
-Bla bla XML upstream, revision, where to put files and so on.
+Repo [#repotool]_ is a tool for using several git repositories in the same
+build. It features a `manifest` format using XML where one defines the git
+repositories and revisions one wants to use. It is possible to include manifests
+in other manifests.
 
-Structure, include other manifests etc.
+Features 101
```
breakreturn: I think this sounds too whimsical. How about Feature overview

docs/articles/baseplatform/repo-manifests.rst
breakreturn: If we could very easily use the real file without maintenance overhead, I'd vote for a shortened example designed to not be maintenance heavy.

breakreturn (changes requested):

tobsan:
@breakreturn @rpannek updated after your comments.

docs/articles/baseplatform/repo-manifests.rst
```diff
@@ -1,6 +1,67 @@
...
+build. It features a `manifest` format using XML where one defines the git
+repositories and revisions one wants to use. It is possible to include manifests
+in other manifests.
 
-Structure, include other manifests etc.
+Feature overview
+----------------
+* Download several git repos simultaneously.
+* Use different revisions for different repos.
+
```
rpannek, line 13: Why do you have those empty lines in the list?

rpannek (changes requested):

docs/articles/baseplatform/repo-manifests.rst
tobsan: Because that's how nested lists works in rst. You also need the empty line before the nest otherwise it thinks I'm doing a definition list (dl tag)

rpannek (approved):

breakreturn (approved):

PR #23, opened 2017-10-06
Add chapter for reproducing yocto builds

BassemMohsen:


docs/articles/baseplatform/reproducing-yocto-builds.rst
```diff
@@ -0,0 +1,91 @@
...
+
+This document describes how to:
+
+* Configure a ``Yocto`` build so that it can be reproduced.
+* Reproduce such a ``Yocto`` build. The reproduction requires that additional artifacts are archived for the Yocto build that is to be reproduced.
+
+Intro to Yocto build reproducibility
+------------------------------------
+
+A bitbake recipe specifies where the source code of a package should be downloaded from. Because there are several ways to do this, a recipe might not specify a specific version but rather the head of some branch or a "latest"-release. Even though this is not recommended, there is no guarantee that developers will not write recipes this way. This of course affects reproducibility as the maintainer of some upstream project may update the head of the branch or the "latest"-release between builds so that the same version of a recipe yields different results. In addition, in cases when reproducibility of a Yocto build is required, the actual availability of the upstream source code packages might cause a problem, e.g. if a tar-archived is moved to some other URL.
```
mardy: "...the same version of a recipe..." -> "...the same recipe compiled at different times..."

"tar-archived" -> "tar archive" (unless you meant something else?)

docs/articles/baseplatform/reproducing-yocto-builds.rst
```diff
@@ -0,0 +1,91 @@
...
+
+The general solution to guarantee reproducibility is to let the Yocto build environment create tar-archives of the downloaded sources and place these in a local mirror during a build. This local mirror can then be used when reproducing the build, a solution that works in most cases. The exception is when a package's source code is stored in a git repository and the recipe for the package does not point to a specific commit but a branch or tag.
```
mardy: It's not clear to me why this wouldn't work: if Yocto makes a tar archive of the git repo, than it cannot change, can it?

docs/articles/baseplatform/reproducing-yocto-builds.rst
```diff
@@ -0,0 +1,91 @@
...
+
+.. code-block:: none
+
+    Function failed: Network access disabled through BB_NO_NETWORK (or set indirectly due to use of BB_FETCH_PREMIRRORONLY) but access requested with command git ls-remote http://git.projects.genivi.org/ipc/common-api-dbus-runtime.git refs/heads/2.1.6 refs/tags/2.1.6^{} (for url None)
+    ERROR: Task 715 (/mnt/ssd/yocto/the-reproduced-build/sources/meta-pelagicore/recipes-core/common-api/common-api-c++-dbus_2.1.6.bb, do_fetch) failed with exit code '1'
+
+The poky layers does not allow recipes that cannot be built with ``BB_NO_NETWORK`` set to "1", but other upstream layers might. If such a recipe is found in some other bitbake layer repository, a bbappend file overriding the ``SRC_URL`` variable can be created so that the recipe instead will point to a specific commit.
```
mardy: "does not" -> "do not"

docs/articles/baseplatform/index.rst
```diff
@@ -7,6 +7,7 @@ Base platform
     prerequisites.rst
     using-the-repo-tool.rst
     repo-manifests.rst
+    reproducing-yocto-builds
```
GordanM: You mean reproducible Yocto build?

docs/articles/baseplatform/reproducing-yocto-builds.rst
```diff
@@ -0,0 +1,91 @@
...
+
+    BB_GENERATE_MIRROR_TARBALLS = "1"
+
+During the Yocto build process, packages are downloaded and stored in the downloads/ directory (or whatever directory has been specified in build/conf/local.conf) of the Yocto build's root directory. When set to "1", ``BB_GENERATE_MIRROR_TARBALLS`` forces bitbake to create tar-balls for all the sources that are fetched using git. Note that the build must be done from scratch otherwise some source code packages may not be packaged into tar-balls. In other words, the build cache needs to be wiped clean before the build is started. When everything is set up, start the build as usual with bitbake.
+
+
+Saving the repo manifest
+------------------------
+
+The above instructions describe how to create the local mirror. To be able to reproduce a build, the manifest file used must also be saved. In practice, this means that the manifest file must be pushed to a git repository because the repo tool used when initializing build directories only fetches manifests using git. Additional benefits of version controlling manifests are covered in ...
```
mardy: I suggest removing the last period, unless you now know what link you wanted to add.

docs/articles/baseplatform/reproducing-yocto-builds.rst
```diff
@@ -0,0 +1,91 @@
+Reproducing yocto builds
+========================
+
+This document describes how to:
+
+* Configure a ``Yocto`` build so that it can be reproduced.
```
GordanM: Configuring a reproducible Yocto builds.

docs/articles/baseplatform/reproducing-yocto-builds.rst
```diff
@@ -0,0 +1,91 @@
+Reproducing yocto builds
```
GordanM: "Reproducible Yocto builds"

docs/articles/baseplatform/reproducing-yocto-builds.rst
```diff
@@ -0,0 +1,91 @@
...
+
+
+Saving the repo manifest
+------------------------
+
+The above instructions describe how to create the local mirror. To be able to reproduce a build, the manifest file used must also be saved. In practice, this means that the manifest file must be pushed to a git repository because the repo tool used when initializing build directories only fetches manifests using git. Additional benefits of version controlling manifests are covered in ...
+If the manifest used only specifies specific commits for each project then the manifest file can be pushed directly to a release branch. If the manifest on the other hand points to some branch or tag of one or more projects, then a manifest with the exact commits used needs to be generated. This can be done with the repo tool from the root of the Yocto build directory after the repo sync command has been issued:
```
mardy: remove "used" from "manifest used".

docs/articles/baseplatform/reproducing-yocto-builds.rst
```diff
@@ -0,0 +1,91 @@
+Reproducing yocto builds
+========================
+
+This document describes how to:
```
GordanM: I would remove this initial paragraph and the bullet points.

docs/articles/baseplatform/reproducing-yocto-builds.rst
```diff
@@ -0,0 +1,91 @@
...
+
+    $ repo manifest -o - -r >> example-manifest.xml
+
+The generated file is then pushed to the relevant release branch of the manifest repository.
+
+
+Reproducing the Yocto build
+---------------------------
+
+The instructions in this section describes how a Yocto build may be reproduced given that the instructions in the above sections have been followed. See Archiving the build environment of a release build for information on how to use the same vagrant machine to host the build.
```
mardy: "describes" -> "describe"

"given" -> "assuming"

Do you need a link for "Archiving the build environment of a release build"?

docs/articles/baseplatform/reproducing-yocto-builds.rst
```diff
@@ -0,0 +1,91 @@
...
+    └── source-mirror/
+
+When initializing the build directory, use a manifest that has been previously associated with the release to be reproduced:
+
+.. code-block:: bash
+
+    $ repo init -u ssh://git@git.example.net/example-group/example-manifest-repository.git -b releases -m example-manifest.xml
+    $ repo sync
+
+Then follow instructions to initialize the build directory.
```
mardy: follow "the" instructions

mardy (changes requested):

docs/articles/baseplatform/reproducing-yocto-builds.rst
```diff
@@ -0,0 +1,91 @@
+Reproducing yocto builds
+========================
+
+This document describes how to:
+
+* Configure a ``Yocto`` build so that it can be reproduced.
+* Reproduce such a ``Yocto`` build. The reproduction requires that additional artifacts are archived for the Yocto build that is to be reproduced.
```
GordanM: Maybe remove the whole paragraph.

docs/articles/baseplatform/reproducing-yocto-builds.rst
```diff
@@ -0,0 +1,91 @@
...
+
+Intro to Yocto build reproducibility
+------------------------------------
+
+A bitbake recipe specifies where the source code of a package should be downloaded from. Because there are several ways to do this, a recipe might not specify a specific version but rather the head of some branch or a "latest"-release. Even though this is not recommended, there is no guarantee that developers will not write recipes this way. This of course affects reproducibility as the maintainer of some upstream project may update the head of the branch or the "latest"-release between builds so that bitbaking the same version of a recipe yields different results. In addition, in cases when reproducibility of a Yocto build is required, the actual availability of the upstream source code packages might cause a problem, e.g. if a tar archive is moved to some other URL.
```
tobsan: Please don't write lines longer than something reasonable, like 80 chars.

tobsan: I think the point about using latest (I assume this refers to AUTOREV) should be in its own paragraph.

docs/articles/baseplatform/reproducing-yocto-builds.rst
```diff
@@ -0,0 +1,91 @@
...
+
+The general solution to guarantee reproducibility is to let the Yocto build environment create tar archives of the downloaded sources and place these in a local mirror during a build. This local mirror can then be used when reproducing the build, a solution that works in most cases. The exception is when a package's source code is stored in a git repository and the recipe for the package does not point to a specific commit but a branch or tag.
```
tobsan: It actually works for branches but not for tags (because tags resolv on the server iirc)

docs/articles/baseplatform/reproducing-yocto-builds.rst
```diff
@@ -0,0 +1,91 @@
...
+
+Building with BB_NO_NETWORK
+---------------------------
+
+Setting the ``BB_NO_NETWORK`` variable to "1" in build/conf/local.conf disables network access for bitbake during the build so that source code can only be fetched from specified local mirrors. This is a good way to guarantee that only specific sources will be used. However, this also means that bitbake will fail when issuing certain commands, for instance git ls-remote, and that it is therefore important to use the ``SRC_REV`` variable instead of pointing to a tag or branch in git-based bitbake recipes. A typical failure could look like this:
```
tobsan: Again, branches shouldn't be an issue, only tags.

docs/articles/baseplatform/reproducing-yocto-builds.rst
```diff
@@ -0,0 +1,91 @@
...
+
+    BB_GENERATE_MIRROR_TARBALLS = "1"
+
+During the Yocto build process, packages are downloaded and stored in the downloads/ directory (or whatever directory has been specified in build/conf/local.conf) of the Yocto build's root directory. When set to "1", ``BB_GENERATE_MIRROR_TARBALLS`` forces bitbake to create tar-balls for all the sources that are fetched using git. Note that the build must be done from scratch otherwise some source code packages may not be packaged into tar-balls. In other words, the build cache needs to be wiped clean before the build is started. When everything is set up, start the build as usual with bitbake.
+
+
+Saving the repo manifest
+------------------------
+
+The above instructions describe how to create the local mirror. To be able to reproduce a build, the manifest file used must also be saved. In practice, this means that the manifest file must be pushed to a git repository because the repo tool used when initializing build directories only fetches manifests using git.
```
tobsan: Actually, to be able to really reproduce a build, one needs to store the static manifest and all the layer sources together with the mirror tarballs. Perhaps that's only relevant in case of a long time release, but otherwise the risk is that something is overwritten or moved in the layers / manifests. Could be worth mentioning.

docs/articles/baseplatform/reproducing-yocto-builds.rst
```diff
@@ -0,0 +1,91 @@
...
+If the manifest only specifies specific commits for each project then the manifest file can be pushed directly to a release branch. If the manifest on the other hand points to some branch or tag of one or more projects, then a manifest with the exact commits used needs to be generated. This can be done with the repo tool from the root of the Yocto build directory after the repo sync command has been issued:
```
tobsan: Release branches or not is not really discussed anywhere. Should we really have it here?

docs/articles/baseplatform/reproducing-yocto-builds.rst
```diff
@@ -0,0 +1,91 @@
...
+The build directory should look like this:
+
+.. code-block:: none
+
+    .
+    ├── build/
+    ├── source-mirror/
+    └── sources/
+
+Once the recipes have been fetched and the environment set up, add the following lines to build/conf/local.conf:
```
tobsan: Once one gets to here, the recipes have already been fetched and the environment set up, no?

tobsan (changes requested):

docs/articles/baseplatform/reproducing-yocto-builds.rst
```diff
@@ -0,0 +1,91 @@
+Reproducing yocto builds
+========================
+
+This document describes how to:
+
+* Configure a ``Yocto`` build so that it can be reproduced.
+* Reproduce such a ``Yocto`` build. The reproduction requires that additional artifacts are archived for the Yocto build that is to be reproduced.
+
+Intro to Yocto build reproducibility
```
GordanM: Reproducible Yocto builds?

docs/articles/baseplatform/reproducing-yocto-builds.rst
GordanM: There are quite a few spelling mistakes. One thing that is visible is "latest"-release, why are the quotation marks used here? And the hyphen?

docs/articles/baseplatform/reproducing-yocto-builds.rst
GordanM: Works for branches.

docs/articles/baseplatform/reproducible-yocto-builds.rst
```diff
@@ -0,0 +1,90 @@
+Reproducible Yocto builds
+=========================
+
+Intro
+-----
+
+A bitbake recipe specifies where the source code of a package should be downloaded from. Because there are several ways to do this, a recipe might not specify a specific version but rather the head of some branch or a latest release.
```
tobsan: This does not make sense to have first in the intro. Shouldn't the intro be about why and a broad overview of how, rather than why there can be issues with this?

Of course, the issue with AUTOREV should be mentioned somewhere, but perhaps not in the introduction.

docs/articles/baseplatform/reproducible-yocto-builds.rst
```diff
@@ -0,0 +1,90 @@
...
+Even though this is not recommended, there is no guarantee that developers will not write recipes to use ``AUTOREV`` to point to latest source revision.
```
tobsan: These lines are still super long. Try staying below 100, or even better, below 80 chars per line.

tobsan (changes requested):
Apart from the malplaced info on AUTOREV and the long lines, this looks really good!

docs/articles/baseplatform/reproducible-yocto-builds.rst
```diff
@@ -0,0 +1,130 @@
...
+=========================
+
+Intro
+-----
+
+This document describes how to configure a Yocto build so that it can be reproduced,
+and how to reproduce such build.
+The reproduction requires that additional artifacts are archived.
+
+A bitbake recipe specifies where the source code of a package should be downloaded from.
```
tobsan: This still doesn't fit in the Intro section. How about putting all of this AUTOREV stuff into a new header called "The problem with AUTOREV" or something like that?

tobsan (changes requested):

tobsan (approved):

docs/articles/baseplatform/reproducible-yocto-builds.rst
```diff
@@ -0,0 +1,134 @@
...
+command has been issued:
+
+.. code-block:: bash
+
+    $ repo manifest -o - -r >> example-manifest.xml
+
+Reproducing the Yocto build
+---------------------------
+
+The instructions in this section describes how a Yocto build may be reproduced given that the
```
mardy: Still unfixed:
"describes" -> "describe"

"given" -> "assuming"

docs/articles/baseplatform/reproducible-yocto-builds.rst
```diff
@@ -0,0 +1,134 @@
...
+head of the branch or the latest release between builds so that bitbaking the same version of
+a recipe yields different results.
+In addition, in cases when reproducibility of a Yocto build is required, the actual availability of
+the upstream source code packages might cause a problem, e.g. if a tar archive is moved to some
+other URL.
+
+The general solution to guarantee reproducibility is to let the Yocto build environment create tar
+archives of the downloaded sources and place these in a local mirror during a build.
+This local mirror can then be used when reproducing the build, a solution that works in most cases.
+The exception is when a package's source code is stored in a git repository and the recipe for the
```
mardy, line 31: I still don't understand this: "It's not clear to me why this wouldn't work: if Yocto makes a tar archive of the git repo, than it cannot change, can it?"

mardy (changes requested):

docs/articles/baseplatform/reproducible-yocto-builds.rst
BassemMohsen: It's quite clear Mardy, Yocto will create the tartballs of the git repos and they won't change.
Yocto won't create a tarball if the recipe have URL point to git repo and SRCREV is AUTOREV.

because Yocto knows that Internet access always needed to retrieve the latest release from the remote git repo.
This behavior is clarified in line 40 and 41, and showing potential error messages you will get when that happens.

docs/articles/baseplatform/reproducible-yocto-builds.rst
BassemMohsen: Agree on "describes" -> "describe"
don't agree on "given" -> "assuming"

docs/articles/baseplatform/reproducible-yocto-builds.rst
mardy: Ah, now it's clear :-) But it's not written as clearly, not even in lines 40 and 41: I suggest you add this to the end of this paragraph (line 32): "...point to a specific commit but a tag: in that case, Yocto won't create a tarball."

docs/articles/baseplatform/reproducible-yocto-builds.rst
mardy: "given" is actually wrong here, because it means that the fact which follows is taken for certain (see also http://www.oxfordlearnersdictionaries.com/definition/english/given-that).
If you don't like "assuming that", a probably better option is "provided that".

Some more opinions: https://ell.stackexchange.com/a/34636

mardy (approved):

PR #26, opened 2017-10-17
Fix bugs in tag extension

demonicblue:
Fix tag metadata beeing overwritten between runs.
Purge tag metadata when documents containing tags are changed
or removed.
Force update of documents containing taglist directives.

### DIFF
--- a/docs/exts/taglist.py
+++ b/docs/exts/taglist.py
@@ -38,7 +38,11 @@ from sphinx.environment import NoUri
 
 
 def initialize(app):
-    app.builder.env.tags = dict()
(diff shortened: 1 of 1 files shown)


tobsan (approved):

mardy (commented):
There is an extra "a" in the commit message. Other than that, I don't know Sphinx, but the changes you are proposing lokk sane, from a Python point of view :-)

demonicblue:
Fixed typo in commit messange and PL message.

mardy (approved):

PR #27, opened 2017-10-18
Tag howto and processes articles

demonicblue:
Create new summary page to list all articles tagged with howto or processes.

I'm not sure which documents are classified as howto and processes but this is a start.
Suggestions?

### DIFF
--- a/docs/articles/licensing/index.rst
+++ b/docs/articles/licensing/index.rst
@@ -21,3 +21,5 @@ tooling in a future proof way. As SPDX is the licensing standard that
 compliance tooling will be using in the future, and indeed already is
 used by tool makers such as WindRiver and Black Duck, they are included
 both for support for tooling and future proofing the format.
+
+.. tags:: process
--- a/docs/index.trst
+++ b/docs/index.trst
@@ -19,3 +19,11 @@ Revision: |release|
     articles/templates/index.rst
 
     articles/workflow/git-workflow.rst
+
+.. toctree::
+    :caption: Categories:
+    :maxdepth: 1
+
+    categories/howto.rst
+
+    categories/process.rst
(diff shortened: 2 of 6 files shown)


mardy:
LGTM, but I'll let someone more familiar with the SWF to give the final approval.

docs/index.trst
```diff
@@ -19,3 +19,11 @@ Revision: |release|
...
 
     articles/workflow/git-workflow.rst
+
+.. toctree::
+	:caption: Categories:
+	:maxdepth: 1
+
+	categories/howto.rst
+
+    categories/processes.rst
```
tobsan: Is this supposed to be indented like this? Shouldn't it be part of the toctree?

tobsan (commented):
Looks good, just had one comment

docs/index.trst
demonicblue: No, you are correct. I also just noticed this and fixed it.
Looks like my editor used tabs instead of spaces.

docs/articles/licensing/index.rst
```diff
@@ -21,3 +21,5 @@ tooling in a future proof way. As SPDX is the licensing standard that
 compliance tooling will be using in the future, and indeed already is
 used by tool makers such as WindRiver and Black Duck, they are included
 both for support for tooling and future proofing the format.
+
+.. tags:: processes
```
breakreturn: I think it should be singular "process" instead of processes. E.g. it's currently howto, not howtos and the tag makes more sense in singular form since it refers to the nature of one article.

breakreturn (changes requested):

docs/articles/licensing/index.rst
demonicblue: Fixed, changed the tag to "process".

tobsan (approved):

breakreturn (approved):

PR #28, opened 2017-10-18
Add chapter on Yocto caching

tobsan:
The chapter describes yocto cache in general, and how it works over HTTP
and NFS.


docs/articles/infrastructure/ci-cd/yocto-cache.rst
```diff
@@ -0,0 +1,101 @@
...
+to be set: ``SOURCE_MIRROR_URL`` and ``SSTATE_MIRRORS``. The first one points to
+where source packages can be found, the second points to the cached binary
+builds.
+
+.. code-block: bash
+    :caption: site.conf
+
+    INHERIT += "own-mirrors"
+    SOURCE_MIRROR_URL ?= "http://example.org/yocto-cache/downloads/"
+    SSTATE_MIRRORS ?= "file://.* http://example.org/yocto-cache/sstate-cache/PATH"
```
rpannek: What does the `file://.*` do here?

docs/articles/infrastructure/ci-cd/yocto-cache.rst
```diff
@@ -0,0 +1,101 @@
...
+        }
+    }
+
+When one updates the cache, this means writing to /var/www/yocto-cache in some
+way.
+
+Cache over NFS
+--------------
+Setting up a cache over NFS requires more setup on the server-side, but on the
+other hand, the cache can be kept private from the outside world.
```
rpannek: You can keep the http also either behind a VPN or even easier with a basic-auth password in the URL: `https://user:password@example.com/yocto-cache/` if you use https the URL is send encrypted so no problem having the password there either.

I'd remove the statement about having it private and would rather add something about everyone being able to update the cache or something like that.

rpannek (changes requested):

docs/articles/infrastructure/ci-cd/yocto-cache.rst
tobsan: It's from the guide on how to set these things up actually. Not sure it is needed actually, but just copied from the existing `site.conf` in pelux-manifests.

docs/articles/infrastructure/ci-cd/yocto-cache.rst
tobsan: Fair point!

tobsan:
Updated!

docs/articles/infrastructure/ci-cd/yocto-cache.rst
```diff
@@ -0,0 +1,101 @@
+Yocto cache
+===========
```
breakreturn: This looks like a how-to basically. I think the title and file name should reflect that something is to be done, e.g. set up a yocto cache, and there should be a label that indicates what type of article it is (see recent additions of labels).

breakreturn (changes requested):

docs/articles/infrastructure/ci-cd/yocto-cache.rst
jeena: Ah yeah, I remember now putting it there and when I tested without then it wouldn't work for some reason.

rpannek (approved):

docs/articles/infrastructure/ci-cd/yocto-cache.rst
tobsan: Fair point.

tobsan:
@breakreturn Updated, please review.

breakreturn (approved):

PR #35, opened 2017-10-24
Fix duplicate label warning

demonicblue:
Replace include directives with toctree. Otherwise the *yocto-project-integration* label is seen twice and generates a warning.

### DIFF
--- a/docs/articles/templates/index.rst
+++ b/docs/articles/templates/index.rst
@@ -14,9 +14,14 @@ build system. The CMake build system can be installed by following the
 instructions on the `CMake install webpage`_ or by using your Linux
 distribution's package manager.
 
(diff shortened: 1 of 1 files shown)


demonicblue:
This PR requires some discussion. Should the same change be applied to all files that are included by the 'include' directive?
Does this impair the ability to use files in a toctree?

tobsan (changes requested):
I think we have discussed earlier on ways to get around this by using toctree instead of include for most of these things. Could you try that instead?

demonicblue:
Yes I can but it will change the structuring of the page somewhat.

tobsan (approved):

breakreturn (approved):

PR #36, opened 2017-10-24
Fix 'not included' warnings

demonicblue:
Add the orphan metadata field in order to avoid sphinx
warning about 'document isn't included in any toctree'.

### DIFF
--- a/docs/articles/baseplatform/creating-a-new-layer.rst
+++ b/docs/articles/baseplatform/creating-a-new-layer.rst
@@ -1,3 +1,5 @@
+:orphan:
+
 Creating a new layer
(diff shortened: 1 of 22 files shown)


breakreturn:
Does this have any impact on the using projects? E.g. will they need to remove all orphan fields?

demonicblue:
As long as the orphan metadata is placed before the document title it will not be rendered.
Reference: http://www.sphinx-doc.org/en/1.4.8/markup/misc.html (also verified this behaviour)
So, no.

tobsan (approved):

breakreturn (approved):

PR #39, opened 2017-11-07
Various fixes in formatting, typos and so on.

tobsan:
Separated into individual commits for each logical change, of course. This is the result of a proof-reading I did.


rpannek (approved):

docs/articles/baseplatform/creating-sdk.rst
```diff
@@ -1,5 +1,7 @@
 :orphan:
 
+.. _building-the-sdk:
```
demonicblue, line 3: What is this used for?

demonicblue (commented):
Looks good, just one little comment.

docs/articles/baseplatform/creating-sdk.rst
tobsan: Valid point! It will be referenced from (for example) the PELUX Baseline SWF. Perhaps we should try to make sure we have labels for all major headings in the blueprint to make it possible to reference them in other docs?

demonicblue: Isn't it enough to use the path and name of the target document for creating hyperlinks?

tobsan: That would be possible, but one wants to link to a specific section of a file using <pre>:ref:\`reference-id\`</pre> style references I think.

demonicblue: Ah, ok.

demonicblue (approved):

GordanM (approved):

PR #43, opened 2017-11-13
Add how-to-oss doc

DunderRoffe:
Signed-off-by: Viktor Sjölind <viktor.sjolind@pelagicore.com>

### DIFF
--- a/docs/articles/oso/checklist-for-making-software-open-source.rst
+++ b/docs/articles/oso/checklist-for-making-software-open-source.rst
@@ -1,0 +1,38 @@
+:orphan:
+
+Checklist for making software open source
+=========================================
+
+.. tagslist:: howto
+
+When making software open source there are some steps that need to be completed.
+The purpose of these steps is to ensure that the code has a sound license, the
+source code repository is understandable and that there are clear descriptions
+on how to contribute. Following is a listing of these steps:
+
+#. Pick an open source license [#owning_copyright_required]_.
+#. Make sure that the license of all dependencies of the software fit the
+   desired license.
+#. Make sure that the following files are present in the root of the source code
+   repository:
+    * LICENSE: Containing the desired license for the code
+    * README.md: Containing information on where to find the documentation and
+      simple getting started info.
+    * CONTRIBUTING.md: Containing information on how to contribute to the
+      project and what rules there are.
+#. Go through file headers of all the source code and make sure that there is a
+   copyright notice, a reference to the LICENSE file and a SPDX tag
+   [#owning_copyright_required]_.
+#. Send a request to make the software open source to the Open Source Office
+   (OSO). The request should contain an accessible link to the source code
+   repository, a statement of what the desired license is as well as a list
+   of business use cases the software needs to fulfill.
+#. The OSO makes sure that the desired license fits the business use cases.
+#. The OSO scans the code to make sure that it is compliant with the desired
+   license.
+#. The OSO registers that the software has been open sourced using the desired
+   license in the OSO decision log.
+#. The code is ready to be pushed to a public repository.
+
+.. [#owning_copyright_required] In order to change the license of a peice of
+   source code, owning the copyright of said source code is required.


docs/articles/oso/how-to-open-source-software.rst
```diff
@@ -0,0 +1,28 @@
+:orphan:
+
+How to open source software
+===========================
+
```
jeremiah: I'd prefer to call this "How to publish open source software." Otherwise, it seems like the active verb is "open" like you're opening a can of beans. But in many cases, the software is already open, at least the dependencies or OS it runs on since we tend to work on GNU/Linux. Also, many readers are going to be dealing with pushing some code that they may not have written from behind the firewall to the open and I think having the document describe that use case is important and likely the most common.

docs/articles/oso/how-to-open-source-software.rst
```diff
@@ -0,0 +1,28 @@
...
+.. tagslist:: howto
+
+When open sourcing source code there are some steps that needs to be completed.
+The purpose of these steps are to ensure that the code has a sound license, the
```
jeremiah: Don't we want to say "Publishing source code has some best practices associated with it along with license obligations."  These two points would cover, I believe, our practice of adding a "CONTRIBUTING.md" as well as demonstrate our understanding that many of the licenses we use *require* "complete and corresponding" source code. At the end of the day a CONTRIBUTING.md file is optional while a LICENSE is not. Shouldn't we firmly ground our HowTo in our understanding of compliance and praxis?

docs/articles/oso/how-to-open-source-software.rst
```diff
@@ -0,0 +1,28 @@
...
+source code repository is understandable and that there are clear descriptions
+on how to contribute. Following is a listing of these steps:
+
+#. Make sure that the following files are present in the root of the source code
+   repository:
+    * LICENSE: Containing the desired license for the code
```
jeremiah: Ensuring a license is there is meaningless if it doesn't correspond with copyright. The first, and maybe most important step, is to ensure that there is copyright authority. Something similar to Debian's "copyright" file for example. We can of course do something different, but we need to know the code has copyright, the copyright is correctly assigned, and the copyright holder allows publication. The license is, of course, the key to the last part, but it needs to be grounded in copyright since it is meaningless on its own.

breakreturn (changes requested):
Looks good to me, but I think we should add a step to make sure there is copyright stated in each file as well.

docs/articles/oso/how-to-open-source-software.rst
```diff
@@ -0,0 +1,28 @@
...
+    * CONTRIBUTING.md: Containing information on how to contribute to the
+      project and what rules there are.
+#. Go through file headers of all the source code and make sure that there is a
+   reference to the LICENSE file and an SPDX tag.
+#. Send a request to open source the code to the Open Source Office (OSO). The
+   request should contain an accessible link to the source code repository as
+   well as a statement of what the desired license is.
+#. The OSO makes sure that the desired license fits the business use cases.
+#. The OSO scans the code to make sure that it is compliant with the desired
+   license.
```
jeremiah: Can we sync with the checklists we use internally?

docs/articles/oso/how-to-open-source-software.rst
```diff
@@ -0,0 +1,28 @@
+:orphan:
+
+How to open source software
+===========================
+
+.. tagslist:: howto
+
+When open sourcing source code there are some steps that needs to be completed.
```
caybro: that _need_ to be ...

docs/articles/oso/how-to-open-source-software.rst
caybro: _is_ to ensure...

docs/articles/oso/how-to-open-source-software.rst
```diff
@@ -0,0 +1,28 @@
...
+#. Make sure that the following files are present in the root of the source code
+   repository:
+    * LICENSE: Containing the desired license for the code
+    * README.md: Containing information on where to find the documentation and
+      simple getting started info.
+    * CONTRIBUTING.md: Containing information on how to contribute to the
+      project and what rules there are.
+#. Go through file headers of all the source code and make sure that there is a
+   reference to the LICENSE file and an SPDX tag.
+#. Send a request to open source the code to the Open Source Office (OSO). The
```
GordanM: I think the context here is missing. It cannot be obvious for a reader what is an "Open Source Office". Is it some well known institution that everyone can refer to? 

It is the community manager in your company or the description on what it is should exist.

caybro (changes requested):
Some typos/grammar

docs/articles/oso/how-to-open-source-software.rst
breakreturn: I agree this is missing, but that might be difficult to add as part of this story. Unless we do a complete placeholder with a very brief description.

It's part of the current backlog in  a project to add a description of OSO and the roles and activities around that.

jeremiah (changes requested):
I'd like to see this be closer to the steps we've outlined in the OSO.

docs/articles/oso/how-to-open-source-software.rst
DunderRoffe: Yes we should

docs/articles/oso/how-to-open-source-software.rst
DunderRoffe: Yes you are correct. Although this line states the point of the actual LICENSE file, which is to state the desired license. If there is copyright majority etc is up to the OSO while scanning etc. See stages below.

DunderRoffe: Also we should add a step which makes sure there are copyright notices in the files...

docs/articles/oso/how-to-open-source-software.rst
DunderRoffe: sure

docs/articles/oso/how-to-open-source-software.rst
DunderRoffe: You are correct in that some is optional some is not etc. However, from our side I would say both are still mandatory. So I do not think it matters to much to state why some are more relevant than others.

Since we are talking about a source code repository the point of complete and corresponding source code breaks down a bit. This is not a publication of a binary, we are just publishing the actual source code. You are indeed correct in that such text is required somewhere although I do not feel that this is the right context. For instance I have another PR about publishing binaries.

DunderRoffe:
Good catch @breakreturn working on it...

DunderRoffe:
@jeremiah @GordanM @breakreturn I changed the filename which unfortunately resolved all of your comments... Would you care to comment again? Also please read through any response I gave above (click show outdated)

docs/articles/oso/how-to-open-source-software.rst
breakreturn: Yeah keep in mind that this is intended to be a checklist to use when opening up code. There needs to be many more texts describing some principles and processes, but we want a How-To to have high cohesion and low coupling for maintenance purposes. Essentially a How-To should be referenced from "above" abstraction wise and almost never the other way around.  think it's OK to reference the OSO since that's difficult to get away from and will always be part of the Blueprint.

breakreturn: I think we should consider if this is really a PELUX Baseline SWF topic. The exact checklist is likely to be different in using projects.

docs/articles/oso/how-to-open-source-software.rst
breakreturn: I think the title and file name is too broad for what this should be. At least if the intention is to have this being the checklist (of a How-To kind of document), then it should be named more as such, e.g. "Checklist for OSS repo peer review". "How to open source software" is a topic on its own that will include processes, roles, principles, other How-To's etc.

breakreturn (changes requested):
Had some comments. Please note that I believe this how-to is PELUX Baseline material rather than Blueprint.

docs/articles/oso/how-to-open-source-software.rst
jeremiah: I like 'checklist for OSS repo peer review'

DunderRoffe:
Rewrote it as a checklist and went through and incorporated the steps outlined in OSO.
@breakreturn is it OK if I keep this PR open til all issues are resolved (without merging) and then move it to SWF for the baseline?

breakreturn:
@DunderRoffe Sure!

DunderRoffe:
@jeremiah I would like your feedback on the changes :)

docs/articles/oso/checklist-for-making-software-open-source.rst
```diff
@@ -0,0 +1,38 @@
...
+   repository, a statement of what the desired license is as well as a list
+   of business use cases the software needs to fulfill.
+#. The OSO makes sure that the desired license fits the business use cases.
+#. The OSO scans the code to make sure that it is compliant with the desired
+   license.
+#. The OSO registers that the software has been open sourced using the desired
+   license in the OSO decision log.
+#. The code is ready to be pushed to a public repository.
+
+.. [#owning_copyright_required] In order to change the license of a pice of
```
jeremiah: s/pice/peice/

jeremiah (approved):
I would still prefer a title like 'Checklist for publishing open source software' but that shouldn't be a blocker.

breakreturn:
I'm also not a fan of the current title or filename. It's a bit broad. Isn't this a checklist when peer reviewing a repo before publishing it? The "process" of making code open source will be much more in the SWF and will, among other things, include the peer review step. At that point it would be good to reference a concise checklist for the peer review. Keeping the separation will allow easier maintenance of the checklist e.g. changes would not affect the process description as much. For this reason the name might be sub-optimal IMHO.

DunderRoffe:
@jeremiah I am worried that using the word "publishing" might cause confusion between putting source code on github (which this is about) and hosting binaries publicly (which this is not about)

@breakreturn The way this text is written it is more the stated process. Peer reviewing is another thing which I weaved in to this list. However, this list also includes the steps executed by the open source office which is not part of the peer review. The reason I include the steps for the OSO is that I think it is important to know what the OSO does and also the OSO can look here to remember what to do. Does that make sense?

jeremiah:
I don't share your concern with binaries vs. source code and the term publishing, but how about we use something like "Checklist for distributing open source software"? This is the same terminology that the GPL uses for example. Also, the verb "making" is too broad as Joakim states. It implies that we're talking about the production of software as opposed to merely the publishing / distributing step.

breakreturn:
@DunderRoffe I see some smaller pieces that aren't checklist stuff I guess, but I'd say in principle it's a checklist. I propose to break out things in order to have a very concise "checklist only" document of the how-to type, and move the instruction-like material to another document.

@jeremiah I don't agree this would be our checklist for distributing software since that will be different between source and binaries for example. And distributing could also be when we release, which this is also not describing.

For maintenance reasons and structure in the SWF, we need a very crisp separation of concerns as often as possible.

jeremiah:
I don't see how this would be different between source and binary except for the added step of building the source code into a binary. All the prerequisites are the same, with the possible exception of archiving for as long as the life of the binary. Oh, and shipping the toolchain with GPL'd code. (Okay, I think I'm starting to make your point @breakreturn) Sigh.

So, "Checklist for publishing source code"?

breakreturn:
@jeremiah :)  Right. Just as an FYI, the way I look at it is that we have more complex goals that one would like to achieve, in the sense that it's not suitable to try do define the goal in one single checklist. So I'm hoping that if a set of steps are common for two goals but one goal include some additions, we could re-use the common part between them. And tie it together on a level  "above" as Instructions.

I believe this will create pretty unnatural structures from time to time, but so far the biggest complexity seems to be with maintenance. I'd rather have an SWF that is slightly difficult to read from an arbitrary entrypoint than a SWF that is incorrect. This is what drives my approach :)

docs/articles/oso/checklist-for-making-software-open-source.rst
```diff
@@ -0,0 +1,38 @@
...
+   repository:
+    * LICENSE: Containing the desired license for the code
+    * README.md: Containing information on where to find the documentation and
+      simple getting started info.
+    * CONTRIBUTING.md: Containing information on how to contribute to the
+      project and what rules there are.
+#. Go through file headers of all the source code and make sure that there is a
+   copyright notice, a reference to the LICENSE file and a SPDX tag
+   [#owning_copyright_required]_.
+#. Send a request to make the software open source to the Open Source Office
```
tobsan, line 26: This assumes there is an OSO available. In that case we should describe what that is, either here, or somewhere else and link to it.

tobsan (changes requested):

docs/articles/oso/checklist-for-making-software-open-source.rst
breakreturn: It should not be described in a how-to like this. I guess we could add a placeholder article with a brief description for now if we feel that's needed. I agree it is slightly strange to refer to something that doesn't exist yet, but it's also pretty difficult to get away from in these situations where we are adding content from many directions. I know @jeremiah will add content about OSO soon though. If we just have a small article added now, we can change that later.

DunderRoffe:
Need to rethink this. I am Closing this PR and trying another structure

PR #45, opened 2017-11-14
Use autosectionlabel extension

tobsan:
Instead of manually creating labels, the autosectionlabel extension
creates labels from each header in the documentation.


DunderRoffe (commented):
It is a bit weird to not be able to have two headings with the same name in different chapters for instance "Introduction".

Can not decide if this is good or bad... 
I kinda like the forced verbosity. Although it is really bad from a layout perspective.

tobsan:
@DunderRoffe it is possible if we add the extra flag that the ref has to be prepended with document name.

demonicblue:
If we prepend refs with document names, doesn't that defeat the purpose of references being decoupled from document names and paths?

tobsan:
@demonicblue yeah, it kinda does. Well not paths, but document names for sure.

tobsan:
@dunderroffe @demonicblue should I interpret this discussion as if we don't want autosectionlabel?

DunderRoffe:
I would say so. Adding the labels are not that annoying anyway. It feels a bit sad although I guess it is better to keep it the way it is

PR #50, opened 2017-11-20
Adding initial intro to Licensing and Compliance section.

jeremiah:
Signed-off-by: Jeremiah C. Foster <jeremiah@jeremiahfoster.com>


DunderRoffe (changes requested):
I am working on text for the oso as well. We should align on structure and what goes where. I like the content but some of it would be even better as introduction to oso in general etching

docs/articles/licensing/index.rst
```diff
@@ -1,8 +1,18 @@
 :orphan:
 
-Licensing
+Licensing and Compliance
 =========
 
+PELUX is open source software. This means that PELUX is built on
```
breakreturn, line 6: What parts of PELUX are intended to be included in this statement? We have a nomenclature that makes "PELUX" a bit ambiguous since it's an offering which is a bit abstract. The OSS parts that we usually describe in these docs are the PELUX Baseline (i.e. the yocto based linux platform), PELUX Software Factory Platform Blueprint (these docs), and the PELUX Baseline Software Factory (the reference SWF deployment that uses these docs).

There's a page for the nomenclature coming and I'm not saying all of the above should be documented here. But maybe we should state here that we mean the "Platform and the Platform Software Factory"? That should be detailed enough to understand roughly what parts we're talking about, without going all in on the nomenclature. And later we could add a reference to the nomenclature and structure descriptions.

breakreturn (changes requested):

docs/articles/licensing/index.rst
tobsan: I think that the way this text is written, it is general enough to imply any or all parts of PELUX. And since all parts of PELUX are open source, the exact nomenclature isn't really an issue?

Perhaps an easy way to mitigate the problem is to say that "all parts of PELUX" are open source?

tobsan (approved):

docs/articles/licensing/index.rst
DunderRoffe: Well that only applies to the baseline right? @tobsan

breakreturn:
@tobsan All parts of PELUX are not open source and some aren't even easy to understand what that would translate to. Referring to things as "PELUX" when we mean other things is currently a major source of confusion in various areas so I'd rather see us being correct on the details than not when we can.

tobsan:
@breakreturn I think we should differentiate between things that **are** PELUX, and things that **build on top** of PELUX. As far as I know, everything that **is** PELUX is open source. But we can of course enumerate all parts if we want to be explicit about it.

breakreturn:
@tobsan Well you are wrong, all parts of PELUX are not open. And keep in mind that I didn't propose that we enumerate all the things, I proposed that we explicitly mention the platform and the platform SWF. It's not the full list of nomenclature etc but at least it is not as unnecessarily unclear as saying "PELUX".

jeremiah:
This is a really inefficient way of working on documentation. I think we need to find another approach.

PR #57, opened 2017-12-07
Change repo url within git workflow instruction

chuanjin:
Signed-off-by: Chuan Jin


tobsan (changes requested):
I think it would be nice to keep cloning over ssh as it makes pushing easier. But we should probably describe somewhere that one needs to have ssh keys set up in order for that to work. Can you change it to add a section on that instead?

chuanjin:
Yeah, sure!

PR #65, opened 2017-12-19
add article on general CI/CD considerations

tobsan:
This is more of an initial article, more content should obviously follow once we dive more into this topic.


docs/articles/infrastructure/ci-cd/considerations.rst
```diff
@@ -0,0 +1,23 @@
...
+* Make build status visible
+* Make sure latest build is available to users
+
+Yocto builds
+------------
+Yocto jobs generally takes a long time to run, and use lots of disk space. To
+mitigate running time, consider :ref:`storing old builds in a cache
+<setting-up-yocto-cache>` in a way that yocto can access. That way, the time of
+a full build is reduced from 4 hours to about 30 minutes. Consider running parts
+of the Yocto jobs in parallel on different machines, if possible.
```
rpannek, line 23: It should have a tag `process` or something like that.

docs/articles/infrastructure/ci-cd/considerations.rst
```diff
@@ -0,0 +1,23 @@
...
+General strategy
+----------------
+Continuous integration and continuous delivery jobs serve the purpose of quick
+turn-around time for developers, providing easy feedback on work in a controlled
+environment. As such, our jobs are designed with some basic properties:
+
+* Build on-commit
+* Fail builds early
+* Make build status visible
+* Make sure latest build is available to users
```
rpannek, line 15: Build or build artifacts? What about nightly builds?

rpannek (changes requested):

docs/articles/infrastructure/ci-cd/considerations.rst
tobsan: Sure but it's not really a process, is it? I'm open for other tags though.

docs/articles/infrastructure/ci-cd/considerations.rst
tobsan: This talks about the broader sense where the build could be a deployed web page or a full yocto build or anything really. The latest should be readily available anyhow.

docs/articles/infrastructure/ci-cd/considerations.rst
rpannek: I think we have process, instruction and howto?

breakreturn: My intention from the beginning was to use the tags to map the content of the SWF with the project setup things (that are non-public i know :P), and this particular article wouldn't necessarily match any of those, but maybe "instruction"... But I don't think we need to tag all content, and we should also ofc consider extending the amount of tags. We can still have a subset of the tags used for the mapping i referred to.

breakreturn (approved):
I think this is a good starting point. It's a big topic and there's a lot of more potential content. But I assume the idea here is not to be complete, but rather add a first piece of the puzzle.

tobsan:
@breakreturn indeed, this is just a starting point.

PR #66, opened 2017-12-21
housekeeping: Fix spelling in all documents

jonte:
This uses the newly added spellchecker to fix all spelling errors found. I removed the dictionary added with the spellchecker and went through the spellchecking manually using `check_spelling.sh -o`.


jonte:
Hmm.. Some words appear to come from a local dictionary on my machine, which are not present in the server dictionary. I will see how I can disable this.

jonte:
Hmm.. On a closer investigation, it seems like my local aspell dictionary (the system dictionary) contains more words than the system dictionary on the build slave, such as "GitHub".. I will just add the missing words for now.

GordanM (approved):

docs/articles/infrastructure/ci-cd/jenkins.rst
```diff
@@ -60,7 +60,7 @@ Then you can open this jenkins instance in your browser and put this password th
 
    http://localhost:8080/jenkins/
 
-After that you can also set up a Nginx instance as a proxy so you don't need the port number in the URL, but you don't need to do that on your development machine.
+After that you can also set up a NGINX instance as a proxy so you don't need the port number in the URL, but you don't need to do that on your development machine.
```
tobsan: I think the preferred spelling here is `nginx`, at least that's how it's written on their own home page.

docs/articles/intro/create-a-swf-deployment.rst
```diff
@@ -25,12 +25,12 @@ probably with the name of the project in it like: "example-software-factory".
     git commit -m "Initial commit"
     git push -u origin master
 
-Add SWF Platform Blueprint as submodule
-=======================================
+Add SWF Platform Blueprint as sub module
```
tobsan: Since we talk about git here, the general term used is `submodule`

docs/articles/sdk/configure-qtcreator.rst
```diff
@@ -5,7 +5,7 @@ Configuring QtCreator to use the SDK
 
 After :ref:`the installation of the SDK <installing-sdk>` and :ref:`sourcing the
 environment <sourcing-the-environment>`, there are a couple of steps needed to
-run QtCreator in the SDK environment. This roadmap assumes that the SDK and
+run QtCreator in the SDK environment. This road map assumes that the SDK and
```
tobsan: I do believe that the term actually is `roadmap` without space.

docs/articles/templates/directory-structure.rst
```diff
@@ -6,7 +6,7 @@ should follow a common directory structure. The tree structure below
 shows a repository with two components sharing a 'doc' directory and
 'component-test' directory. Many repos will likely contain only one
 component, in which case the structure should be the same but with only
-one subdirectory with the source code etc.
+one sub directory with the source code etc.
```
tobsan: Shouldn't this then be `sub-directory`?

tobsan (changes requested):
Nice work, had comments on some specific changes only.

docs/articles/infrastructure/ci-cd/jenkins.rst
jonte: You're probably right.. I was looking at this: https://www.nginx.com/resources/wiki/ <-- But https://nginx.org/en/ uses all lower case.

tobsan (approved):

PR #67, opened 2017-12-27
architecture: Populating the chapter structure

GordanM:
Signed-off-by: Gordan Markuš <gordan.markus@pelagicore.com>


breakreturn:
I'm not sure I see how you want the new 'architecture' tag to be used. The idea with the tags is to allow a reader to browse by tag as a compliment to having an index tree structure. For me this means that all articles that have a tag should be fairly usable on their own, either stand-alone or as good entry points. So I think we should watch out for trying to duplicate what can already be found from the index tree and be a bit sparse on what we tag so we don't accidentally create noise for the user. Not saying there's a problem with this commit as such but please keep it in mind.

For example, I don't think it's nice with what we have today already that an article named 'Introduction' shows up as a Process, when it's just an intro and entry-point to one specific process. It's not clear cut how the tags should be used so it's a case by case thing and requires some evaluation, but in this case I don't see how a reader is helped by a generic "Introduction" showing up as a Process. This is what I'd like us to avoid.

GordanM:
@breakreturn noted! I've missed this design decision. My line of thought was that every group should have a tag - obviously I was wrong. Will fix.

breakreturn (approved):

breakreturn:
@GordanM Yeah it's not a clearly stated design decision so not easy to know about. There's actually no guidelines currently on how to do this, the whole meta part if WiP.

PR #68, opened 2017-12-27
Change structure for more clarity

breakreturn:
Signed-off-by: Joakim Gross <joakim.gross@pelagicore.com>


breakreturn:
To make it possible to have a better separation of the "meta" of the SWF structure and how to use the Blueprint etc, from any SWF deployments, I'd like to have these articles gathered in a different way. This is to prepare for more content but I'm separating the structural change from the content.

GordanM (approved):

PR #71, opened 2018-01-03
architecture: Capitalize subtitles

GordanM:
Signed-off-by: Gordan Markuš <gordan.markus@pelagicore.com>


tobsan:
Is this the way it is done in most other chapters? If so, go ahead (but I think it is actually bad style). If not, we should keep the existing style.

breakreturn (changes requested):
This is not how it's done in general in the SWF docs.

GordanM:
Yup, just checked it. The arc42 template confused me as they do it this way. I'll use this PR to fix the caps on all the Architecture related chapters.

GordanM:
Updated @breakreturn and @tobsan .

tobsan (approved):

breakreturn (approved):

PR #76, opened 2018-01-15
Booting QEMU: Add note about adapting rootfs path when using custom i…

erikboto:
…mage

Signed-off-by: Erik Botö <erik.boto@pelagicore.com>


GordanM (approved):

rpannek:
The headline of the commit message looks really weird because it's too long. We're following https://wiki.gnome.org/Git/CommitMessages/ which says:

> First line (the brief description) must only be one sentence and should start with a capital letter unless it starts with a lowercase symbol or identifier. Don't use a trailing period either. It's recommended to not exceed 50 characters but it's not always possible/easy to follow this limit. In any case, don't exceed 72 characters.

tobsan (approved):

tobsan (approved):

PR #85, opened 2018-01-31
sde: Add more details on how to extend the SDE

rpannek:
This patch adds more info about how to extend the SDE, how to add
add more and different tests and adds the introduction page to the
sidebar of the blueprint.

It also removes the word PELUX from in front of all the mentions
of the SDE, this way it's easier to reuse the text in different
projects.

### DIFF
--- a/docs/articles/sde/introduction_to_sde.rst
+++ b/docs/articles/sde/introduction_to_sde.rst
@@ -3,11 +3,11 @@
 Software Development Environment
 ********************************
 
-PELUX Software Development Environment (SDE) is a tool that creates a virtual
+The Software Development Environment (SDE) is a tool that creates a virtual
 machine with all the tools needed when developing towards a PELUX platform.
 More specifically the SDE is setup using vagrant to generate a VirtualBox
-instance. In order to facilitate this, the PELUX SDE requires an instance of
-PELUX SDK to be present in the root directory (same level as Vagrantfile) while
+instance. In order to facilitate this, the SDE requires an instance of
+the SDK to be present in the root directory (same level as Vagrantfile) while
 provisioning the VM.
 
 Starting the SDE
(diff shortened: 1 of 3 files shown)


GordanM (approved):

docs/articles/sde/introduction_to_sde.rst
```diff
@@ -17,16 +17,17 @@ Dependencies:
 * Vagrant
 * VirtualBox
 
-To start the PELUX SDE VM, go to root directory (the same directory as the
-Vagrantfile is located in) the SDE run
+To start the SDE VM, check out the repository, put the SDK into it's rood
```
thenor: Typo on "root" (misspelled as "rood")

thenor (changes requested):
There is a typo on the second line of the commit message ("add" twice")

rpannek:
Ok, fixed that, @thenor

docs/articles/sde/introduction_to_sde.rst
```diff
@@ -35,13 +36,24 @@ Following is a list of environment variables used by vagrant when provisioning
 the SDE.
 
 * SDK_FILE_NAME: Name of the self-extracting SDK package to install into the
-  SDE. Can include wildcards. Defaults to `oecore*toolchain*sh`.
+  SDE. Can include bash wildcards. Defaults to `oecore*toolchain*sh`.
```
DunderRoffe, line 39: Might be out of scope but we should probably change the default value to `pelux*toolchain*sh`

DunderRoffe (commented):
I think we should mention that this is developed using TDD. So when adding new stuff we assume that people also add tests for that as well.

docs/articles/sde/introduction_to_sde.rst
rpannek: Hm yeah, or even "*toolchain*sh" in the script but also change the sdk-name from oecore... to pelux... I'll create a issue for that.

rpannek: https://github.com/Pelagicore/meta-pelux/issues/90

rpannek:
Ok, fixed the TDD thing.

PR #88, opened 2018-02-07
Change spell-checker to use a sphinx plugin

tobsan:
Only downside so far was the need for a custom version string filter


docs/python_modules/enchantfilters/SWFFilters.py
```diff
@@ -0,0 +1,30 @@
+from enchant.tokenize import Filter
```
demonicblue, line 19: What about copyright and license statements?

demonicblue (changes requested):

docs/python_modules/enchantfilters/SWFFilters.py
tobsan: good point, will add!

docs/python_modules/enchantfilters/SWFFilters.py
```diff
@@ -0,0 +1,48 @@
...
+#
+# For further information see LICENSE
+
+from enchant.tokenize import Filter
+import string
+
+def all_match(string, allowed_chars):
+    return all(c in allowed_chars for c in string)
+
+class GitCommitFilter(Filter):
```
demonicblue, line 25: This doesn't look like it's used. Is it for a future feature?

docs/python_modules/enchantfilters/SWFFilters.py
```diff
@@ -0,0 +1,48 @@
...
+    """ If a word looks like a git commit hash (long or short), ignore it """
+
+    def _skip(self, word):
+        allowedchars = string.hexdigits.lower()
```
demonicblue: This should be a class var since it's always the same.

docs/python_modules/enchantfilters/SWFFilters.py
tobsan: Yeah actually I thought I was gonna use it but I ended up not needing it. I just figured it might be useful so I let it be.

docs/python_modules/enchantfilters/SWFFilters.py
tobsan: Good point!

demonicblue (commented):
I'm seeing that the build is failing because of missing dependencies. Should we just add pip install commands to the jenkins file?

tobsan:
@demonicblue no I will add the deps to the server actually. I have a thing in progress to run it all in a docker container, but that will be later.

tobsan:
@demonicblue build works now. Had some issues with pip dependencies...

demonicblue (approved):

docs/CMakeLists.txt
```diff
@@ -47,16 +47,16 @@ configure_file(
     "${BINARY_BUILD_DIR}/conf.py"
     @ONLY)
 
-add_custom_target(spell-check
-    COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/check_spelling.sh" "-q" -w "${CMAKE_CURRENT_SOURCE_DIR}"
```
rpannek, line 51: Will this still be able to handle different lists for the blueprint and the project using it?

tobsan:
@rpannek I have a pattern for how to work with several wordlists now, but I can't show it unless this gets merged first :P

tobsan:
I also opened an issue on the spelling packages' GitHub page: https://github.com/sphinx-contrib/spelling/issues/10

docs/CMakeLists.txt
tobsan: Not out of the box, but for the PELUX SWF I have a working solution for it, and there is an open PR to add it to the speller plugin.

rpannek: Now with breaking changes release notes would be really nice ;)

rpannek (approved):

docs/CMakeLists.txt
tobsan: Not out of the box, but for the PELUX SWF I have a working solution for it, and there is an open PR to add it to the speller plugin.

tobsan: I added an instruction in the README for how to do this in a project that uses the blueprint.

PR #89, opened 2018-02-08
Jenkinsfile: build the docs in docker instead

tobsan:
Removes the requirement to keep up with dependencies on the build
servers, which is nice, plus the Dockerfile acts as documentation for
what's needed to build the docs.


tobsan:
Thanks @rpannek for the Dockerfile!

Jenkinsfile
```diff
@@ -8,7 +8,7 @@
 
 pipeline {
     agent {
-        node { label 'Sphinx' }
+        dockerfile { label 'DockerCI' }
```
demonicblue: Looks good but you can specify the label for nodes supporting Docker in the global jenkins settings instead.  _Manage Jenkins_ -> _Configure System_ -> _Docker Label_

demonicblue (commented):
Looks good. Just a small comment.

Jenkinsfile
tobsan: Updated.

demonicblue (approved):

PR #94, opened 2018-02-19
SWFFilters.py: allow X.Y.Z-Codename-GitRevision

tobsan:
The codename (such as alpha, beta, or possibly pyro) was not allowed
before, causing the spell check to crash on the PELUX Baseline SWF.


docs/python_modules/enchantfilters/SWFFilters.py
```diff
@@ -46,8 +47,18 @@ class VersionStringFilter(Filter):
...
-        (version,sep,revparse) = word.partition('-')
+        (version,sep,rest) = word.partition('-')
         if version == word:
             return False # probably not a version string then
         else:
-            return all_match(version, self.versionchars) and all_match(revparse, self.allowedchars)
+            # The version string has to look right
+            if not all_match(version, self.versionchars):
+                return False
+            else:
```
demonicblue: Try removing this for easier reading

docs/python_modules/enchantfilters/SWFFilters.py
```diff
@@ -46,8 +47,18 @@ class VersionStringFilter(Filter):
     versionchars = string.digits + "."
 
     def _skip(self, word):
-        (version,sep,revparse) = word.partition('-')
+        (version,sep,rest) = word.partition('-')
         if version == word:
             return False # probably not a version string then
         else:
```
demonicblue: Try removing this for easier reading

demonicblue (commented):
I find the nesting a bit hard to read. Could it be a good idea to remove the else-statements or would it diverge from the pythonic way?

docs/python_modules/enchantfilters/SWFFilters.py
tobsan: I kept this one actually but I removed the others.

demonicblue (approved):

PR #95, opened 2018-02-19
Should be tags in article

chuanjin:
To be picked up by index, we should use tags rather than taglist, I assume?


tobsan:
I think so yes! But I would like @demonicblue to have a say as well.

docs/articles/sde/how-to-cross-compile-using-qtcreator-in-sde.rst
```diff
@@ -2,7 +2,7 @@
 
 How to cross compile using Qt Creator and PELUX SDE
 ===================================================
-.. taglist:: howto
+.. tags:: howto
```
demonicblue, line 5: It should be `.. tag:: howto`

demonicblue (changes requested):

docs/articles/sde/how-to-cross-compile-using-qtcreator-in-sde.rst
chuanjin: But if you look at other articles, there is "tags" used, since you can use comma to separate multiple tags, right?

demonicblue: The tag directive is used to tag articles. The taglist directive is used to list articles matching a certain tag.

demonicblue: My bad! It was the documentation for the plugin which was wrong.

demonicblue (approved):

docs/articles/sde/how-to-cross-compile-using-qtcreator-in-sde.rst
chuanjin: Great we found a bug in the doc! :)

tobsan (approved):

PR #98, opened 2018-02-27
Remove label from jenkins.rst

kursatkobya:
To be able to include jenkins.rst from other files, labels are needed to be removed from the file.

### DIFF
--- a/docs/articles/infrastructure/ci-cd/jenkins.rst
+++ b/docs/articles/infrastructure/ci-cd/jenkins.rst
@@ -1,7 +1,5 @@
 :orphan:
 
-.. _jenkins:
-
 Jenkins
 =======
 


docs/articles/infrastructure/ci-cd/jenkins.rst
```diff
@@ -1,7 +1,3 @@
-:orphan:
```
tobsan: Don't remove the orphan, because without it any other project that doesn't use this file won't build.

docs/articles/infrastructure/ci-cd/jenkins.rst
```diff
@@ -1,7 +1,3 @@
...
-
-.. _jenkins:
```
tobsan: This doesn't seem to be used in the PELUX Baseline, so I guess it's fine.

tobsan (changes requested):

demonicblue:
Why do you need to include it in another page? If this is an issue we should consider labels for all pages under infrastructure.

tobsan:
And I agree with @demonicblue that trying to avoid using includes is the easiest strategy from a sphinx perspective.

kursatkobya:
@demonicblue because we have some project based differences so we should add this file and enhance it with project specific details. 

Please feel free to bring this as an issue for applying this to all pages.

tobsan (approved):

kursatkobya:
As a blueprint repository i believe it make sense to be permissive.
So i do not think that we should prevent other projects to include blueprint files.

PR #100, opened 2018-03-05
Use the term "deploy" instead of "write" for image

erikboto:
Instead of talking about "writing an image" it makes more sense to
talk about deploying an image, especially if additional information
about more exotic targets are added where it's more than just writing
to disk.

Signed-off-by: Erik Botö <erik.boto@pelagicore.com>


thenor (approved):
Looks good. Many builds are failing in Jenkins now, hopefully the failure for this PR is just the same error as for the other PRs.

BassemMohsen (approved):

PR #102, opened 2018-03-13
Update instruction of booting qemu using help script

chuanjin:
Signed-off-by: Chuan Jin <cjin@luxoft.com>


docs/articles/baseplatform/booting-qemu-image.rst
```diff
@@ -10,15 +10,35 @@ virtualization feature if found on the host system and run the virtual
...
+	  -c | --cmdline          	Specify kernel command line. This would allow power users to change vga or root, etc.
+								Defaults to "vga=0 uvesafb.mode_option=1280×720-32 root=/dev/hda console=ttyS0 rw oprofile.timer=1"
+	  -m | --mem              	Maximum amount of guest memory (Size is in megabytes)
+								Defaults to 512
+	  -t | --gdb_port         	The port is forwarded so that gdbserver can listen on it
+								Then gdb can be run from develop machine to connect to the port for remote debugging
+								Defaults to 3333
+	  -d | --dev              	Run QEMU development image instead of normal image
+								Defaults to core-image-ntg-x-dev-qemux86-64.ext4
+	  -s | --static           	Run QEMU development image with static IP (192.168.7.2)
```
tobsan: Is 192.168.7.2 the default or is it hard-coded? I don't like hard-coded values.

docs/articles/baseplatform/booting-qemu-image.rst
```diff
@@ -10,15 +10,35 @@ virtualization feature if found on the host system and run the virtual
...
 
 .. code-block:: bash
 
-    $ sudo apt-get install qemu-kvm libvirt-clients libvirt-daemon-system
+	$ ./qemu_launcher -h
+	usage: scripts/qemu_launcher [options]
+	[options] is any of the following:
+	  -f | --folder           	Specify a source folder of the QEMU image
+								Defaults to tmp/deploy/images/qemux86-64
+	  -i | --hda              	Specify rootfs file. Defaults to core-image-ntg-x-qemux86-64.ext4
```
tobsan: We probably should have the default be `core-image-pelux-minimal-qemux86-64.ext4` or whatever it is called, if anything.

tobsan (changes requested):

docs/articles/baseplatform/booting-qemu-image.rst
chuanjin: Yes, I updated the script in meta-pelux but forgot to update the docs. Good catch!

docs/articles/baseplatform/booting-qemu-image.rst
chuanjin: I think it is better to involve Basem or Jonathan for this discussion, since they know better than me. I remember we have this static IP in our local.conf, but maybe it is needed to upstream to meta-pelux as well.

tobsan: ping @BassemMohsen @jonte

jonte: Looks hardcoded. The project where this script stems from uses this IP (always). I think it would make sense to make the IP configurable, so that this is usable by other projects as well.

BassemMohsen: So the runqemu script relies on the default variables of:
TEST_TARGET_IP = "192.168.7.2"
TEST_SERVER_IP = "192.168.7.1"

Booting the qemu outside testing environment, could be launched to any values.
The instruction for booting with static ip address however give you opportunities to use TAP/TUN to get internet access to the qemu machine. and it doesn't matter changing the default values as long as user know what he's doing like updating default value here:
/etc/systemd/network/10-pelux-ethernet.network

BassemMohsen: The environment variables lives here:
meta/lib/oeqa/runexported.py

docs/articles/baseplatform/booting-qemu-image.rst
```diff
@@ -10,15 +10,39 @@ virtualization feature if found on the host system and run the virtual
 machine in better performance.
 See KVM feature for QEMU [#qemu_kvm_feature]_.
 
-Installing KVM and its dependencies
+QEMU launch script
 -----------------------------------
```
tobsan: This underline is still very long.

docs/articles/baseplatform/booting-qemu-image.rst
```diff
@@ -96,6 +104,17 @@ commands on the host machine:
...
 
+Instead of using the default target IP ``192.168.7.2`` and the server address
+``192.168.7.1``, it is possible to specify the target IP and server IP as shown below:
+
+.. code-block:: bash
+
+   ./qemu_launcher -f FOLDER -i core-image-pelux-minimal-qemux86-64.ext4 -s --target_ip 192.168.11.18 --server_ip 192.168.11.1
+
+Please aware that to make the new static IP work, it is also needed to update
+``STATIC_IP_ADDRESS`` in ``conf/variant/common/local.conf`` in meta layer and
```
tobsan: Let's just say `local.conf` here.

tobsan (approved):

PR #105, opened 2018-03-16
Code change for building swf along with swf-blueprint whenever pull request comes to swf-blueprint

SKumarDas:


Jenkinsfile
```diff
@@ -11,14 +11,28 @@ pipeline {
...
+        string(name: "BP_BRANCH", defaultValue: "master")
+    }
+
     stages {
         stage('Download') {
             steps {
                 // Delete old files
                 sh 'rm -rf .[^.] .??* *'
+		sh "git init"
+		sh  "git remote add jenkins https://github.com/Pelagicore/software-factory.git"
```
tobsan: Shouldn't this use `params.BP_REMOTE` instead?

tobsan (changes requested):

Jenkinsfile
SKumarDas: Thanks for pointing out. Actually I am experimenting with jenkinsfile how it behaves in order to complete my task. once I am done with that I will write a clean code and start pull request.

tobsan:
@SKumarDas Why did you close this PR?

GordanM:
@SKumarDas with gitlab the work flow (or at least how we use it) would be to amend your commit and force push it to the same branch. That way you don't need to have:

- A lot of commits
- A lot of opened/closed pull requests 

NOTE: Try not to force push to the master branch :)

tobsan:
@GordanM I'd say there is nothing wrong with having lots of commits while testing (it's actually nice to be able to track exact changes), but one can later just squash them and force push, just as you suggested.

GordanM:
@tobsan being a reviewer and looking through 50 commits is something I wouldn't be inclined to do. Especially if there were comments. But maybe you have a point :)

tobsan:
@GordanM perhaps it would make sense to mark the PR as WIP for the time being then, but yeah. When reviewing I'd want to keep the number of commits as few as possible, but as many as needed.

SKumarDas:
@GordanM  @tobsan .. My bad, I couldn't get the email notifications to your comments as I am yet to configure an email client. I have opened another pull request with proper comments as that was just a test to get familiar with git , git pull requests and jenkins. 
But you both have good points. I will do it properly next time. Thanks.

PR #106, opened 2018-03-19
Code change for building swf along with swf-blueprint whenever pull request comes to swf-blueprint

SKumarDas:
Hi,
Please review the code and help to merge the code.
Kindly refer to the logs of stages in pipeline of jenkins to check for the validity of the code.
Thank you,
Sapan Das


Jenkinsfile
```diff
@@ -11,14 +11,31 @@ pipeline {
...
+
+		sh "git init || git init" // some times it fails to initialize so trying another time, in case it fails.
+		sh "git remote add jenkins ${REMOTE_SWF_LINK}"
+		sh "git fetch jenkins ${BRANCH_SWF}"
+		sh "git checkout jenkins/${BRANCH_SWF}"
+
+		// Initialize the submodules and fetch the new code change into it
+		
+		sh "git submodule init && git submodule update"
+ 		sh "cd docs/swf-blueprint && git pull ${REMOTE_SWF_BP_LINK} pull/${env.CHANGE_ID}/head"
```
tobsan: This will not work if there is no CHANGE_ID, like when building on master.

Jenkinsfile
```diff
@@ -11,14 +11,31 @@ pipeline {
...
     stages {
         stage('Download') {
             steps {
+
                 // Delete old files
+
                 sh 'rm -rf .[^.] .??* *'
 
-                // Checkout the git repository and refspec pointed to by jenkins
-                checkout scm
```
tobsan: Would it be possible to make these steps conditional on there being a CHANGE_ID?

Imho, it makes sense to do the regular build in regular cases, and the build that this PR introduces only when there is a SWF BP change done (so pull requests, then).

Or does this work properly when building static branches as well, such as master? I'm not sure.

Jenkinsfile
```diff
@@ -11,14 +11,31 @@ pipeline {
...
+		// Get The latest SWF src
+
+		sh "git init || git init" // some times it fails to initialize so trying another time, in case it fails.
+		sh "git remote add jenkins ${REMOTE_SWF_LINK}"
+		sh "git fetch jenkins ${BRANCH_SWF}"
+		sh "git checkout jenkins/${BRANCH_SWF}"
```
tobsan: With these changes, we always build with the SWF Baseline, and never build the BP on its own. We should do both when there are changes to the BP.

tobsan (changes requested):
1) Squash your commits
2) Should this really go into the general Jenkinsfile? Would it make sense to have a separate Jenkinsfile for the test-build with SWF case?
3) What we're missing here is building the SWF BP as a standalone. That is perhaps not very useful for most cases, but it is supported and it should work.

Jenkinsfile
```diff
@@ -11,14 +11,33 @@ pipeline {
         dockerfile true
     }
 
+    parameters {
+        string(name: "REMOTE_SWF_LINK", defaultValue: "https://github.com/Pelagicore/software-factory.git")
+        string(name: "REMOTE_SWF_BP_LINK", defaultValue: "https://github.com/Pelagicore/software-factory-blueprint.git")
```
tobsan: I'm a bit skeptical to this particular parameter. Since this Jenkinsfile is in the BP repo already, it should not be needed. We should be able to do "checkout scm" or similar and then work with that. Perhaps we can check out the code in a subdirectory and copy it later to the SWF build or something?

Jenkinsfile
```diff
@@ -11,14 +11,35 @@ pipeline {
...
+
                 // Delete old files
+
                 sh 'rm -rf .[^.] .??* *'
 
-                // Checkout the git repository and refspec pointed to by jenkins
-                checkout scm
+                // Get The latest SWF src
+
+                sh "git init || git init" // some times it fails to initialize so trying another time, in case it fails.
```
tobsan: You can use this instead
```
retry(2) {
    sh "git init"
}
```

Jenkinsfile
```diff
@@ -11,14 +11,35 @@ pipeline {
...
+                sh "git remote add jenkins ${REMOTE_SWF_LINK}"
```
tobsan: To properly show that the SWF and the Blueprint are built separately, I would prefer if these steps were also inside a `dir() { }` block.

tobsan (changes requested):
Nice work! I like this solution. I had some smaller comments and I have one more general comment: I would like to not combine the build and configure steps for the SWF and BP, but have them separate. So you would do:
```
stage('Configure SWF') { ... }
stage('Build SWF') { ... }

stage('Configure BP') { ... }
stage('Build BP') { ... }
```

To keep them more separated.

tobsan (approved):

PR #107, opened 2018-03-27
Jenkinsfile: break out building of SWF / SWF BP

tobsan:


thenor (approved):

GordanM (commented):
Nitpick - there is a `.` in the commit shortlog.

Jenkinsfile
```diff
@@ -14,90 +32,48 @@ pipeline {
...
-
-        // Configure and build SWF
-        stage('Configure SWF') {
-            steps {
-                script {
-                    String buildParams = "-DENABLE_PDF=OFF"
-                    sh "cd ${SWF_DIR} && cmake -H. -Bbuild ${buildParams}" // Configure for SWF baseline
-                }
+                // It's for testing (BP changes + SWF code)
+                sh "cp -r ${SWF_BP_DIR}/* ${SWF_DIR}/docs/swf-blueprint/"
```
SKumarDas, line 55: Just wanted to check if this builds the swf-blueprint as well along with SWF code whenever you lunch build from SWF. Because there is no submodule init done here. Can you clarify on this ?

tobsan: Yes, the `${SWF_DIR}/docs/swf-blueprint/` directory is empty until one initializes the submodule. Since we don't want the actual submodule, but rather the one from whatever version of the blueprint we're building, we don't need to do `init` but we can just copy the files that we already have from our `checkout scm` step.

tobsan:
@GordanM fixed.

GordanM (approved):

PR #108, opened 2018-03-28
Add article about Private Docker Registry

rpannek:
Sometimes it's really convinient to run your own docker registry in
a project where you can host custom docker images which can be used
by the CI and the developers without the need to build them manually
every time you need them.

This article explains how to set up such a registry and push a
custom image to it.


tobsan (approved):

tobsan:
Noticing a build issue on Jenkins, seems to be unrelated to this.

rpannek:
I re-ran it and see that there is a misspelled word, I'll fix that.

PR #110, opened 2018-04-10
Add article about FreeIPA

demonicblue:
FreeIPA is an open source identity management system. It allows for
fine grained control over which linux systems users can access and what they
can do. In addition it provides an LDAP service for autentication users
in web application and other supported services.

Signed-off-by: Henrik <hhugo@luxoft.com>


rpannek (changes requested):
Please add it to the index so one can reach it while building the blueprint: https://github.com/Pelagicore/software-factory-blueprint/blob/master/docs/articles/infrastructure/ci-cd/index.trst

This is also why the build fails.

docs/articles/infrastructure/ci-cd/freeipa.rst
```diff
@@ -0,0 +1,238 @@
+FreeIPA
```
rpannek, line 3: Please add `:orphan:` at the top so that people can choose not to pull it in.

rpannek (changes requested):

docs/articles/infrastructure/ci-cd/freeipa.rst
demonicblue: Done. I don't think we need to add this article to any index in the blueprint, right?

docs/articles/infrastructure/ci-cd/freeipa.rst
```diff
@@ -0,0 +1,240 @@
+:orphan:
+
+FreeIPA
+=======
+
+FreeIPA is an identity management system which comes with an LDAP server, kerberos and a web UI built in. It has the ability to "Manage Linux users and client hosts in your realm from one central location with CLI, Web UI or RPC access." It can thus be used to centrally manage accounts and allow users to login to multiple servers and websites with the same credentials.
```
tobsan: This and many other lines are too long, please wrap them around 80 chars.

docs/articles/infrastructure/ci-cd/freeipa.rst
```diff
@@ -0,0 +1,240 @@
...
+
+    hostname ipaserver.example.com
+
+Then set up ``/etc/hosts`` file with the fully qualified domain name pointing to the servers IP:
+
+.. code-block:: none
+    :caption: Example /etc/hosts
+
+    # Replace with your own values
+    198.51.100.5 ipaserver.example.com ipaserver
```
tobsan, line 48: If this is a real IP address, please use an example address or one from a reserved block (like 10.0.0.0/8)

docs/articles/infrastructure/ci-cd/freeipa.rst
```diff
@@ -0,0 +1,240 @@
...
+
+    sudo apt-get install freeipa-server freeipa-server-dns
+
+.. code-block:: bash
+    :caption: Setup the FreeIPA server
+
+    # Replace the values below with your own. Never use example.com in a real deployment.
+    ipa-server-install --setup-dns --hostname=ipaserver.example.com --realm=EXAMPLE.COM --domain=example.com --no_hbac_allow 
+
+During installation you will be asked to setup forwarding DNS servers. Either set them to the DNS provided by your organization, internet provider or public ones like Google ``8.8.8.8`` or Cloudflare ``1.1.1.1`` ``1.0.0.1``.
```
tobsan: For completeness, perhaps one should add 8.8.4.4 to the google ones.

docs/articles/infrastructure/ci-cd/freeipa.rst
```diff
@@ -0,0 +1,240 @@
...
+    sudo apt-get install freeipa-client
+
+Set fully qualified hostname with
+
+.. code-block:: bash
+
+    # Example value
+    sudo hostname host.example.com
+
+Edit /etc/resolvconf/resolv.conf/base and put the ipaserver first:
```
tobsan: Set the path in ``backticks``please.

docs/articles/infrastructure/ci-cd/freeipa.rst
```diff
@@ -0,0 +1,240 @@
...
+
+    netgroup:       nis sss
+    sudoers: files sss
+
+If no home directory is created upon loging in with an ipa user, use the commands below. 
+
+.. code-block:: none
+
+    # Source: https://bgstack15.wordpress.com/2017/06/26/enabling-mkhomedir-on-ubuntu-for-freeipa/
+    sed -i -r -e 's/Default:\s\w+/Default: yes/;' /usr/share/pam-configs/mkhomedir
```
tobsan: Perhaps a comment saying that it would replace whatever was written after "Default: " with "yes" would be useful here.

docs/articles/infrastructure/ci-cd/freeipa.rst
```diff
@@ -0,0 +1,240 @@
...
+
+Edit nsswitch.conf and remove references to ldap.
+
+Check if ldap users disappeared from the system:
+
+.. code-block:: bash
+
+    getent passwd
+
+Ldap users will have a higher uid than system users.
```
tobsan: Shouldn't LDAP be written with caps?

tobsan (changes requested):
Looks good, I had mostly wording/formatting comments here.

docs/articles/infrastructure/ci-cd/freeipa.rst
demonicblue: Actually I'm using a block reserved for documentation, see https://tools.ietf.org/html/rfc5737 for example IP addresses. To quote:

>The blocks 192.0.2.0/24 (TEST-NET-1), 198.51.100.0/24 (TEST-NET-2),
>   and 203.0.113.0/24 (TEST-NET-3) are provided for use in
>   documentation.

tobsan: Ah nice, I learned something new today!

tobsan (approved):

docs/articles/infrastructure/ci-cd/freeipa.rst
rpannek: I do think you should add it to the index in the blueprint so when I build the blueprint I can browse to it and look at it.

rpannek (changes requested):

docs/articles/infrastructure/ci-cd/index.trst
```diff
@@ -9,6 +9,7 @@ Continuous Integration
     build-slaves
     howto-yocto-cache
     docker-registry
+    volumerize
```
rpannek: Shouldn't this be freeipa instead?

demonicblue: You're right! I'm getting confused working on multiple documents :s

PR #113, opened 2018-04-23
Add SWF_BP chapter for GPLv3 license in Automotive industry

SKumarDas:
Chapter contents:
1. GPLv3 licensed production software Vs automotive industry
2. Difference between GPLv3 license usage of production and development images
3. How to facilitate blacklisting of GPLv3 licensed software in Yocto
4. Impact on PELUX by removal of GPLv3 licensed software from PELUX

Signed-off-by: Sapan Das <SKumarDas@luxoft.com>

### DIFF
--- a/docs/articles/licensing/gplv3/gplv3-blacklist.rst
+++ b/docs/articles/licensing/gplv3/gplv3-blacklist.rst
@@ -1,0 +1,64 @@
+GPLv3 in automotive
+-------------------
+
+Car makers prioritize safety and security of vehicle software.
+Car makers won't let users modify vehicle software because they carefully test
+and certify their software and any change can compromise safety. GPLv3 software
+explicitly states that users must be able to modify software, and installation
+information provided even if that includes private encryption keys. Car makers
+currently would rather prevent GPLv3 software from being introduced into their
+systems than allow users to modify vehicle software.
+
+Fortunately, Yocto provides a means to screen licenses based on different
+parameters.
+
+Screening GPLv3 licensed software in Yocto
+------------------------------------------
+
+1. Add layer **meta-gplv2**
+2. Remove `initramfs-live-install` and `initramfs-module-install` from
+   environment variable PACKAGE_INSTALL.
+   *(file: core-image-minimal-initramfs.bbappend)*
+
+   .. code-block:: bash
+
+      PACKAGE_INSTALL_remove = "initramfs-live-install initramfs-module-install"
+3. Remove `grub-efi` from WKS_FILE_DEPENDS_BOOTLOADERS for x86-64 arch.
+   *(file: image_types_wic.bbclass)*
+
+   .. code-block:: bash
+
+     WKS_FILE_DEPENDS_BOOTLOADERS_remove_x86-64 = "grub-efi"
+4. Remove `parted` from RDEPENDS_${PN}.
+   *(file: initramfs-module-install-efi_1.0.bb)*
+
+   .. code-block:: bash
+
+     RDEPENDS_${PN}_remove = "parted"
+5. Select `system-boot` as the bootloader instead of default `grub-efi`.
+   *(file: systemd-boot.bbclass)*
+
+   .. code-block:: bash
+
+     EFI_PROVIDER = "systemd-boot"
+6. Blacklist GPLv3 license 
+
+   .. code-block:: bash
+
+     INCOMPATIBLE_LICENSE = "\
+        GPL-3.0 \
+        LGPL-3.0 \
+        AGPL-3.0 \
+      "
+7. Remove `cairo`
+
+   .. code-block:: bash
+
+    PACKAGECONFIG_remove = "cairo"
+
+Impact of removing GPLv3 licensed software
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+SWUpdate feature for Raspberry Pi currently requires `partprobe`, a GPLv3 tool,
+for re-partitioning. So as of now it will be non-functional with the removal of
+`parted` package.
+
--- a/docs/articles/licensing/index.rst
+++ b/docs/articles/licensing/index.rst
@@ -24,4 +24,10 @@ compliance tooling will be using in the future, and indeed already is
 used by tool makers such as WindRiver and Black Duck, they are included
 both for support for tooling and future proofing the format.
 
+.. toctree::
+    :caption: GPLv3 license in automotive industry
+    :maxdepth: 1
+
+    gplv3/gplv3-blacklist.rst
+
 .. tags:: process
--- a/docs/spelling_wordlist.txt
+++ b/docs/spelling_wordlist.txt
@@ -342,3 +342,7 @@ compat
 IPv
 subdomain
 uninstall
+GPLv
+initramfs
+efi
+partprobe


docs/articles/licensing/gplv3/gplv3-blacklist.rst
```diff
@@ -0,0 +1,86 @@
+GPLv3 licensed production software Vs automotive industry
+---------------------------------------------------------
+
+Modern cars have around 100 million lines of code running on them,12 with 70%
```
AshleyNair: the number 12 is a superscript, so it should be aligned properly. Not sure if the problem is here or with the fromatting in git.
As this is a reference note, you also need to provide the link. The link for that is https://www.genivi.org/sites/default/files/BMW_Case_Study_Download_040914.pdf
The link provided in Jeremiah's paper doesn't work as intended.

docs/articles/licensing/gplv3/gplv3-blacklist.rst
```diff
@@ -0,0 +1,86 @@
...
+
+Locking the software to the hardware by signing the original software image
+with a cryptographic key so that only an image provided by the supplier will 
+boot or install – is a common practice in embedded devices.
+This process of signing software images – so only images with the right key 
+will boot or install effectively prevents a user from modifying the software
+on the device since they have no access to the key needed to allow their 
+modified version to boot or install.
+
+With regard to the GPLv3, the difficulty is with only the anti-tivoization
```
FlorentRevest: Maybe we should define tivoization before using it? It's not a word I often hear or use :)

docs/articles/licensing/gplv3/gplv3-blacklist.rst
```diff
@@ -0,0 +1,86 @@
...
+	    LGPL-3.0 \
+	    AGPL-3.0 \
+	    MPL-1.1 \
+	    SGI-1 \
+	    CECILL-2.0 \
+	    CDDL-1.1 \
+	    EPL-1.0 \
+	    CPL-1.0 \
+	"
+	PACKAGECONFIG_remove = "cairo"
```
FlorentRevest: Why is this line in the "Deprecate GPLv3 and other" paragraph?

docs/articles/licensing/gplv3/gplv3-blacklist.rst
GordanM: The software factory is not a blogpost but a documentation so this kind of introduction is not appropirate IMO. We speak with facts and data, if you can't quote a reference for this number I don't see its value.

docs/articles/licensing/gplv3/gplv3-blacklist.rst
```diff
@@ -0,0 +1,86 @@
...
+of that code being in the head unit. Complexity is a non-trivial issue in 
+automotive software design. In addition to being complex, cars can be dangerous.
+
+Locking the software to the hardware by signing the original software image
+with a cryptographic key so that only an image provided by the supplier will 
+boot or install – is a common practice in embedded devices.
+This process of signing software images – so only images with the right key 
+will boot or install effectively prevents a user from modifying the software
+on the device since they have no access to the key needed to allow their 
+modified version to boot or install.
```
GordanM: This paragraph is really unclear for a layman. Basically what should be written here is: What is the implication of GPLv3, why the automotive doesn't want it, how is GPLv2 different. The more you speak in long sentences the more you will confuse someone.

docs/articles/licensing/gplv3/gplv3-blacklist.rst
```diff
@@ -0,0 +1,86 @@
...
+
+With regard to the GPLv3, the difficulty is with only the anti-tivoization 
+clause, and the reason for disfavoring that license is safety. 
+
+References
+^^^^^^^^^^
+
+`<https://www.gnu.org/licenses/gpl-faq.en.html#Tivoization>`_
+`<http://www.ifosslr.org/ifosslr/article/view/102/207>`_
+
```
GordanM: Where is the most important reference? https://www.yoctoproject.org/docs/latest/ref-manual/ref-manual.html#migration-2.3-gplv2-and-gplv3-moves

docs/articles/licensing/gplv3/gplv3-blacklist.rst
```diff
@@ -0,0 +1,86 @@
...
+------------------------------------------------------------------
+
+1. Add layer **meta-gplv2**
+2. Remove "initramfs-live-install" and "initramfs-module-install" from 
+   environment variable PACKAGE_INSTALL.
+   *(file: core-image-minimal-initramfs.bbappend)*
+
+   .. code-block:: bash
+
+      PACKAGE_INSTALL_remove = "initramfs-live-install initramfs-module-install"
```
GordanM, line 25: Add another line break after this line?

docs/articles/licensing/gplv3/gplv3-blacklist.rst
```diff
@@ -0,0 +1,86 @@
...
+^^^^^^^^^^
+
+`<https://www.gnu.org/licenses/gpl-faq.en.html#Tivoization>`_
+`<http://www.ifosslr.org/ifosslr/article/view/102/207>`_
+
+Difference between GPLv3 license usage of production and development images
+---------------------------------------------------------------------------
+
+* The development image can contain GPLv3 softwares even without commercial
+  license. E.g - gdbserver(GPLv3) can be used in development image
```
GordanM: Please use a space between a `word (`

GordanM: Please use backticks for components.

docs/articles/licensing/gplv3/gplv3-blacklist.rst
SKumarDas: @AshleyNair  Sure. Thank you. I will have a look.

docs/articles/licensing/gplv3/gplv3-blacklist.rst
```diff
@@ -0,0 +1,86 @@
...
+	    CECILL-2.0 \
+	    CDDL-1.1 \
+	    EPL-1.0 \
+	    CPL-1.0 \
+	"
+	PACKAGECONFIG_remove = "cairo"
+
+Impact of removing GPLv3 licensed software
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+The software update feature won't work in PELUX, as it uses "parted" tool.
```
FlorentRevest: It is not exactly true. SWUpdate itself doesn't depend on parted. However, the artifacts I generate for the raspberrypi3 use a lua script that repartition the disk and at the end of this script partprobe (from parted) is used to inform the kernel of partitions changes https://github.com/sbabic/meta-swupdate-boards/blob/master/recipes-extended/images/update-image/raspberrypi3/emmcsetup.lua#L76
Maybe we could find an alternative to partprobe that is not GPLv3...

docs/articles/licensing/gplv3/gplv3-blacklist.rst
SKumarDas: @AshleyNair test comment.

docs/articles/licensing/gplv3/gplv3-blacklist.rst
```diff
@@ -0,0 +1,86 @@
+GPLv3 licensed production software Vs automotive industry
+---------------------------------------------------------
+
+Modern cars have around 100 million lines of code running on them,12 with 70% 
+of that code being in the head unit. Complexity is a non-trivial issue in 
+automotive software design. In addition to being complex, cars can be dangerous.
+
+Locking the software to the hardware by signing the original software image
```
TSAnsari: Please add reference numbers to the mentioned references below, for example [1] [2] and use these numbers where ever you have taken the content from the referred articles. If you are using the exact words from these references then they should be clearly placed in the quotes.

docs/articles/licensing/gplv3/gplv3-blacklist.rst
GordanM: Why is cairo here in the base image? What component depends on it?

docs/articles/licensing/gplv3/gplv3-blacklist.rst
```diff
@@ -0,0 +1,86 @@
...
+	    GPL-3.0 \
+	    LGPL-3.0 \
+	    AGPL-3.0 \
+	    MPL-1.1 \
+	    SGI-1 \
+	    CECILL-2.0 \
+	    CDDL-1.1 \
+	    EPL-1.0 \
+	    CPL-1.0 \
+	"
```
GordanM: I have no clue about half of these licenses and I believe that we should have proper comments why they are here? What is the background why wasn't only L/A/GPLv3 removed?

GordanM (changes requested):

docs/articles/licensing/gplv3/gplv3-blacklist.rst
```diff
@@ -0,0 +1,86 @@
...
+* In production image if GPLv3 non-commercial licensed software is used,
+  then the vendor has to provide the necessary tools to change the 
+  software as well.
+
+
+How to facilitate blacklisting of GPLv3 licensed software in Yocto
+------------------------------------------------------------------
+
+1. Add layer **meta-gplv2**
+2. Remove "initramfs-live-install" and "initramfs-module-install" from
```
GordanM: Use backticks for components - `component`.

GordanM (changes requested):

docs/articles/licensing/gplv3/gplv3-blacklist.rst
jeremiah: We'd need a non-GPLv3 component unless partprobe is being used as a tool and not residing on the target. Is that the case?

docs/articles/licensing/gplv3/gplv3-blacklist.rst
jeremiah: SWF might provide some context though @GordanM perhaps. Though I do think we can reduce the verbiage a bit and write something like "The GPLv3's anti-tivoization clause is interpreted as meaning one has to provide installation information to modify source code if you ship GPLv3 licensed software. The installation information requirement means that car makers cannot have an encrypted boot sequence in the vehicle without giving away their private keys. This compromises security and safety in the vehicle and as a consequence, GPLv3 software is often blacklisted by car companies. This documentation describes a workaround to this problem through using older GPLv2 licensed software."

docs/articles/licensing/gplv3/gplv3-blacklist.rst
jeremiah: Would you review my proposed changes above?

jeremiah (changes requested):

docs/articles/licensing/gplv3/gplv3-blacklist.rst
jeremiah: Many of the licenses, if not all, are considered 'problematic' from a OEM's point of view, just not for the same reasons as the GPLv3. One option might be to describe only removing GPLv3 in this section and not focus on other licenses in this part of the SWF.

docs/articles/licensing/gplv3/gplv3-blacklist.rst
GordanM: @jeremiah thanks a lot! That sounds good, as long as @FlorentRevest comment is addressed as well (what is actually tivoization, a short explanation and link would be enough). 

Always the smooth talker :)

docs/articles/licensing/gplv3/gplv3-blacklist.rst
GordanM: Done!

docs/articles/licensing/gplv3/gplv3-blacklist.rst
```diff
@@ -0,0 +1,86 @@
...
+
+1. Add layer **meta-gplv2**
+2. Remove "initramfs-live-install" and "initramfs-module-install" from 
+   environment variable PACKAGE_INSTALL.
+   *(file: core-image-minimal-initramfs.bbappend)*
+
+   .. code-block:: bash
+
+      PACKAGE_INSTALL_remove = "initramfs-live-install initramfs-module-install"
+3. Remove "grub-efi" from WKS_FILE_DEPENDS_BOOTLOADERS for x86-64 arch.
```
GordanM: Backticks instead of " on all components would be nicer :)

docs/articles/licensing/gplv3/gplv3-blacklist.rst
GordanM: Thanks @jeremiah ! That's good to know, I believe we should not address them in this chapter, because, like you have mentioned, there are other reasons not to include those.

docs/articles/licensing/index.rst
```diff
@@ -24,4 +22,10 @@ compliance tooling will be using in the future, and indeed already is
 used by tool makers such as WindRiver and Black Duck, they are included
 both for support for tooling and future proofing the format.
 
+.. toctree::
+    :caption: GPLv3 license in Automotive industry
```
GordanM: Why is Automotive in caps and other nouns not? It should be either all or none, please check with other chapters.

docs/articles/licensing/gplv3/gplv3-blacklist.rst
FlorentRevest: Unfortunately, it is actually part of the target so we need to get rid of it.
I tried a bunch of alternatives:
hdparm -z /dev/mmcblk0
sfdisk -R /dev/mmcblk0
blockdev --rereadpt /dev/mmcblk0
... But none of them would work

docs/articles/licensing/gplv3/gplv3-blacklist.rst
jeremiah: The answer to why a given component is in PELUX should really come from the PELUX Architect.

docs/articles/licensing/gplv3/gplv3-blacklist.rst
jeremiah: If it is necessary for the core functionality of the software, I don't think we should remove it based on license. There are a number of alternatives for dealing with GPLv3, such as additional exceptions and modification under homologation (https://en.wikipedia.org/wiki/Homologation). 

In addition, the point of PELUX is to be an open source starting point for IVI, so we have no issues with GPLv3, in fact we like it since it is a better license than other strong copyleft licenses. While I think it is courteous of us to describe how you can manage the situation where a customer requests GPLv3 removal, it should not block the work we're doing in PELUX.

docs/articles/licensing/gplv3/gplv3-blacklist.rst
SKumarDas: @AshleyNair comment!

rpannek: I find it super weird to on the one hand to say "GPLv3's tivoization clause compromises security and safety" and as a sollution to that to propose some old ass versions of the software which still are under a GPLv2 but are unmaintained and full of security holes, it just doesn't make any sense to me, it's basically lying.

docs/articles/licensing/index.rst
rpannek: It should also be "... in the Automotive ..."

docs/articles/licensing/gplv3/gplv3-blacklist.rst
FlorentRevest: Maybe we could keep partprobe then. For now, it is only needed to inform the kernel of disk changes after re-partitioning the raspberry pi's sdcard. I assume that "in production" we would have pre-partitioned disks that wouldn't require the usage of partprobe, so it's no big deal.

docs/articles/licensing/gplv3/gplv3-blacklist.rst
SKumarDas: @AshleyNair test comment!

jeremiah: You're absolutely right @rpannek, but don't shoot the messenger. :-) 

You surely understand though that a car maker might worry about their public reputation if there is an accident that was caused by user modified software. That is not unreasonable, some companies, like Volvo, base their brand on safety, of course they want to preserve the brand. Disallowing user modification is one way to help ensure safety. I don't think anyone is saying what you put in quotes however and I think that showing other users how to use yocto's license filtering mechanism is *not* a bad thing as it can be used for other purposes.

I am not really sure about this whole enterprise frankly and I think that is what you're pointing to in some way Richard. Why are we doing this? We are completely foss so why are we telling folks how to do a non-foss configuration in our foss product? Shouldn't this be done on a case-by-case basis internally for customers? There we can warn them of the actual side effects (unmaintained, buggy code) and provide other approaches than merely blacklisting.

GordanM: @rpannek and @jeremiah the answer to why are we doing this: because we have done it X number of time and we will probably do it a Y number of times. It will save someone some effort :). Is it good practice from a foss point of view? Noup. Do we agree with it? Not entirely. Is it useful to have this documented? Definitively yes. That is at least my point here :).

breakreturn: I agree with @GordanM, this topic always comes up. And PELUX is to a high degree intended to be used in automotive projects that want to understand these things.

But I do think we should phrase ourselves as dry and static as possible in these docs. I think the net outcomes of discussions and analysis is more important than arguing for why. My design recommendation for the SWF docs have always been to only provide the rationale behind something when it's needed to provide context so the reader/user can understand.

For me this is a matter of "sometimes someone might need to do this" which we know is true in basically all our projects. And then we describe how to do "this". Arguing for why it's good or bad is not a topic for these docs IMO unless it carries a needed context.

jeremiah: Okay, fair enough. But in that case I feel we should pretty much just reproduce the upstream documentation as faithfully as possible, i.e. https://wiki.yoctoproject.org/wiki/How_do_I#Q:_How_do_I_build_an_image_without_GPLv3_Licensed_packages_.3F

frznlogic: I agree to what @breakreturn say. I would personally like to add a comment that this is not an endorsement or "disendorsement" of either habit on our side though.

frznlogic: I agree to what @breakreturn say. I would personally like to add a comment that this is not an endorsement or "disendorsement" of either habit on our side though.

docs/articles/licensing/gplv3/gplv3-blacklist.rst
frznlogic: Late reply, I would disagree to the premise @gordanm. I don't think we are writing for any layman and hence shouldn't dumb it down too much. As has already been discussed though, we should only describe the problem as objective as possible and our proposal for a solution to it.

docs/articles/licensing/gplv3/gplv3-blacklist.rst
frznlogic: Can we refer to some other places that discuss bad/good licenses and their implications?

docs/articles/licensing/gplv3/gplv3-blacklist.rst
GordanM: @frznlogic fair point. Layman might be streched but it might be someone new to the field like the case you are a part of right now. It should be easily understandable and readable, that is the main point :).

docs/articles/licensing/gplv3/gplv3-blacklist.rst
SKumarDas: @GordanM "cairo" package has been removed in earlier practice of removing GPLv3 license. What should we do now? Shall I remove 3g DISTRO FEATURE which brings in python3-pygobject=>python3-pycairo or shall I remove cairo, which will remove python3-pycairo 
anyway?

docs/articles/licensing/gplv3/gplv3-blacklist.rst
```diff
@@ -1,39 +1,10 @@
...
-Locking the software to the hardware by signing the original software image
-with a cryptographic key so that only an image provided by the supplier will 
-boot or install – is a common practice in embedded devices.
-This process of signing software images – so only images with the right key 
-will boot or install effectively prevents a user from modifying the software
-on the device since they have no access to the key needed to allow their 
-modified version to boot or install.
-
-With regard to the GPLv3, the difficulty is with only the anti-tivoization 
-clause, and the reason for disfavoring that license is safety.
```
SKumarDas: This looks cool. Thank you. I am going to integrate this into my changes.

docs/articles/licensing/gplv3/gplv3-blacklist.rst
GordanM: Thanks Sapan for investigating this. For me the real chain of dependencies goes like this: DISTRO_FEATURES+=3g brings in ofono which in turn brings in the python stuff ONLY IF the tests are enabled. The tests can be disabled though by a CMake flag, please check: https://git.yoctoproject.org/cgit.cgi/poky/plain/meta/recipes-connectivity/ofono/ofono.inc. The main question here is why are the tests PACKAGES compiled? Is it a runtime dependency for something else? Or does this recipe automatically compile these tests, since this flag is set in the CMake flag. We should resolve these questions first before we proceed. Otherwise we are shooting in the dark :)

SKumarDas: @GordanM -
1. The test package is part of ofono itself. They are just python files and called during yocto qa API calls. This is a run time dependency.
2.  Tried with "--disable-test" and without any flag as well in CMake flag macro "EXTRA_OECONF", It doesn't help in solving the problem. So CMake doesn't play a role in getting the test package in.
3.  In order to remove python3-pycairo we can remove the "3g" feature as we have done for "bluetooth" in other project.

jeremiah: If the components are only for testing, then they are not required on the target and as such should be ignored. Why spend time and energy on something that will never be installed on the target?

SKumarDas: @jeremiah agreed!

SKumarDas:
@jeremiah @GordanM @FlorentRevest , I have addressed the comments. Please help to merge the pull request.

jeremiah (approved):

docs/articles/licensing/index.rst
GordanM: @SKumarDas please address this comment :)

docs/articles/licensing/index.rst
```diff
@@ -1,5 +1,3 @@
-:orphan:
```
GordanM: Hmm should this be removed though?

docs/articles/sde/how-to-use-dev-image.rst
```diff
@@ -6,7 +6,7 @@ How to use development image
 
 Introduction
 -------------
-The dev image is a superset of its corresponding production image for each
+The dev image is a super set of its corresponding production image for each
```
GordanM: This is an unrelated change to the actual commit and should be either a separate commit or unchanged :)

GordanM (approved):
Added small comments that need to be addressed and then I am fine with merging this.

docs/articles/licensing/index.rst
SKumarDas: @GordanM , This is no more a single orphan page. It contains "GPLv3 in automotive" page as well as. It enables to access GPLv3 related topics from main menu as well.

docs/articles/sde/how-to-use-dev-image.rst
SKumarDas: I will revert this change and make a separate commit then. Thanks.

GordanM: Superset is a noun that is written together http://www.yourdictionary.com/superset. Not sure why would you change it.

docs/articles/licensing/index.rst
frznlogic: But will you _always_ have this page included in sub projects inheriting from the blueprint? Ie, there are currently implemented SWF project documentations right now that will likely break if you remove that :orphan: because the page is not included in the toctree anywhere.

SKumarDas: @frznlogic If that is the case then, the page should be labeled as :orphan: . I will revert that change then. Thanks.

PR #114, opened 2018-04-26
Add a page under SDK blueprints describing the usage of GENIVI DLT

FlorentRevest:
PELUX exposes system logs through GENIVI DLT. It is then interesting to
document how to compile dlt-viewer and how to use it to connect to a
target.

Signed-off-by: Florent Revest <revestflo@gmail.com>


docs/articles/sdk/read-system-logs.rst
```diff
@@ -0,0 +1,99 @@
...
+ .. code-block:: bash
+
+    bin/dlt-viewer
+ 
+Configuring dlt-viewer for your target
+--------------------------------------
+
+dlt-viewer must be configured to point to your PELUX machine. You can do that
+from "*Config -> ECU Add*". On the first tab, named "ECU", feel free to use any
+Id and Description you want and make sure the "Interface Type" is set to TCP.
```
frznlogic: What is the ID and description used for?

docs/articles/sdk/read-system-logs.rst
```diff
@@ -0,0 +1,99 @@
...
+  dlt-daemon
+
+Compiling dlt-viewer
+--------------------
+
+Since dlt-daemon is already present on PELUX, you only need to compile
+dlt-viewer on your development machine. For that, you need to have `git`,
+`cmake` and the Qt SDK installed on your machine.
+
+Then, clone the dlt-viewer repository:
```
frznlogic: Remove Then,

docs/articles/sdk/read-system-logs.rst
```diff
@@ -0,0 +1,99 @@
...
+- **dlt-daemon**: running on PELUX, collecting logs and distributing them on
+  a network
+- **dlt-viewer**: running on a developer machine, accessing distant logs from
+  dlt-daemon
+
+Compiling dlt-viewer
+--------------------
+
+Since dlt-daemon is already present on PELUX, you only need to compile
+dlt-viewer on your development machine. For that, you need to have `git`,
```
frznlogic: I'd prefer a bullet list of the dependencies

docs/articles/sdk/read-system-logs.rst
```diff
@@ -0,0 +1,99 @@
...
+
+    git clone https://github.com/GENIVI/dlt-viewer
+
+Compile it with cmake:
+
+.. code-block:: bash
+
+    mkdir build
+    cd build
+    cmake ..
```
frznlogic: Can you run - -build at the same time as you generate the makefiles? If so we could shorten the build instructions. For example

Cmake -Bbuild - H. - - build 

Instead of all these instructions

docs/articles/sdk/read-system-logs.rst
```diff
@@ -0,0 +1,99 @@
...
+Id and Description you want and make sure the "Interface Type" is set to TCP.
+On the second tab, named "IP", use the IP address of your target and and make
+sure the port is set to 3490. You can then hit the "OK" button and your machine
+should appear in the "Project" part of the main window.
+
+.. note:: If you don't know the IP address of your target, the `Raspberry Pi
+          documentation`_ describes how to find a new device on the network.
+          This can be used for any target, not just Raspberry Pi
+
+.. _`Raspberry Pi documentation`: https://www.raspberrypi.org/documentation/remote-access/ip-address.md
```
frznlogic: We have traditionally listed the links at the end of the document. I'm not sure how it's done in the rest of the blueprint though.

docs/articles/sdk/read-system-logs.rst
```diff
@@ -0,0 +1,99 @@
...
+
+You can use the line edit in the top action bar to search for words such as
+"*set_default*". You can also, enable regular expression search and filter with
+requests such as "*set\_.*_trace_status*"
+
+Logging data
+^^^^^^^^^^^^
+
+On PELUX, DLT comes with a couple of tools that make life easier. One of those
+tools is called `dlt-adaptor-stdin` and it can be used to log data from the
```
frznlogic: Would be nice with a link to where this comes from.

frznlogic (changes requested):
I like the document, just some comments 🙂

docs/articles/sdk/read-system-logs.rst
FlorentRevest: I couldn't find a way to do it with less commands than:
cmake -H. -Bbuild
cmake --build build
Which is already better than what I had wrote but it looks like --build expects the Makefile to be here and we can't use it in one command

docs/articles/sdk/read-system-logs.rst
FlorentRevest: It's just for the UI, I don't think it's really that important.

GordanM:
@FlorentRevest nice work! I was wondering did you try to write a small service using DLT and maybe adding that example or maybe that is redundant because you would get that after building and checking the docs?

FlorentRevest:
I haven't tried but could. Or I could also point to a simple example like this one https://github.com/GENIVI/dlt-daemon/blob/master/examples/example1/example1.c ?

GordanM:
@FlorentRevest no worries, I think it would be redundant. Oh that would be nice!

FlorentRevest:
There ya go. I tried to keep the "Logging data" section short as it's slightly out of scope of "Reading System Logs with DLT" so I just pointed to the relevant sources of information.

GordanM (approved):
LGTM. @frznlogic ? :)

frznlogic (approved):

PR #118, opened 2018-05-31
Fix badly formatted tables in directory-structure

frznlogic:
The directory-structure.rst file contained two tables that where badly
formatted for newer versions of sphinx. Apparently this is stricter with
newer versions of sphinx.

WARNING: Would love some more testing with older versions of sphinx before this is merged ;). 

Signed-off-by: Oscar Andreasson <oscar.andreasson@pelagicore.com>

### DIFF
--- a/docs/articles/templates/directory-structure.rst
+++ b/docs/articles/templates/directory-structure.rst
@@ -40,45 +40,50 @@ directories in the tree structure above. Some directories are only
 relevant in some situations, e.g. a 'test-data' dir is only relevant
 if there is any actual test data of course.
 
-====================================== ===========
+====================================== =======================================================================
 Directory                              Description
-====================================== ===========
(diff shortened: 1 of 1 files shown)


DunderRoffe:
What is old and new version in this context?

DunderRoffe (approved):
I think the newer look is more readable in ASCII which is always preferable IMO

DunderRoffe:
The build issue seems weird

frznlogic:
I've seen issues with building the old code with sphinx 1.8.4, while it seems to work building with 1.7.2 and older versions, but I haven't tested the new code I wrote with 1.7.2 and older versions of sphinx. 

As for the build error, it seems to be a problem with connection refused from docker... @thenor isn't this an old and known issue?

PR #124, opened 2018-06-25
Add reference to git ssh configuration

fhajredini:


sashko (approved):

docs/articles/baseplatform/prerequisites.rst
```diff
@@ -7,9 +7,7 @@ Yocto requires the following packages to be installed in order to build an image
 
 .. code-block:: bash
 
-    git sed wget cvs subversion git-core coreutils unzip gawk python-pysqlite2
-    diffstat help2man make gcc build-essential g++ chrpath libxml2-utils
-    libsdl1.2-dev texinfo python3 gcc-multilib g++-multilib
+    $ apt-get install git sed wget cvs subversion git-core coreutils unzip gawk python-pysqlite2 diffstat help2man make gcc build-essential g++ chrpath libxml2-utils libsdl1.2-dev texinfo python3 gcc-multilib g++-multilib
```
thenor: I think the purpose of these lines was just to list the dependencies, not to give the command for installing them

docs/articles/intro/create-a-swf-deployment.rst
```diff
@@ -128,4 +133,5 @@ The content of the files can be copied and adapted from the PELUX Baseline
 Software Factory which is one such SWF Deployment. The README.md should
 contain a description on how to build the project with CMake.
 
+.. _`Git workflow`: http://pelux.io/software-factory/master/swf-blueprint/docs/articles/workflow/git-workflow.html#general-setup
```
thenor: I'm wondering if the might be another way to refer to a chapter within the documentation. I think there is some way of tagging and referring to the tag.

thenor (changes requested):
The pre-requisites should not be written in the way you changed it to. Might be a better way. And if needed add a ref to the "install" block.

docs/articles/baseplatform/prerequisites.rst
fhajredini: Check the code block for debian. Its listed as command (apt-get install), yet seperated in 3 lines. I can fix it either way as long as they are consistent but this made more sense.

docs/articles/intro/create-a-swf-deployment.rst
fhajredini: Yes. Will fix it now

docs/articles/baseplatform/prerequisites.rst
sashko: I see what you mean. The thing is the list is very Debian/Ubuntu specific, since in packages names may be different for other distros.

sashko: @fhajredini as long as it looks nice on the page, either way, is fine.

thenor: Since this is not actually a code, but a list of deps I think @fhajredini will change it to a list in some way. Right?

fhajredini: Yeap, im doing it now.

thenor (approved):

PR #125, opened 2018-06-27
Update sde intro article

TSAnsari:
Signed-off-by: Tariq Ansari <tansari@luxoft.com>


docs/articles/sde/introduction_to_sde.rst
```diff
@@ -16,6 +16,9 @@ Dependencies:
 
 * Vagrant
 * VirtualBox
+* Self-extracting SDK package to install into the SDE. `See, Building the SDK installer`_.
+
+.. _`See, Building the SDK installer`: http://pelux.io/software-factory/master/swf-blueprint/docs/articles/baseplatform/creating-sdk.html#building-the-sdk
```
sashko: See

sashko (changes requested):

sashko (approved):

docs/articles/sde/introduction_to_sde.rst
fhajredini: Linking internal pages should use references. Instead of giving the link, you can just reference with :ref:`building-the-sdk` (the ` sign is converting the text to code)

fhajredini (changes requested):
Linking internal pages should use references. Instead of giving the link, you can just reference with :ref:`building-the-sdk` (the ` sign is converting the text to code)

fhajredini (approved):

PR #129, opened 2018-07-03
Configure QtCreator: Add a explanation for replacing CFLAGS/CXXFLAGS

sparkleholic:
Configure QtCreator: Add a explanation for replacing CFLAGS/CXXFLAGS

'CC' or 'CXX' environment variable of the sdk-terminal could indicate
not only the compiler executable path but also default compiler flags
which are not specified in 'CFLAGS' or 'CXXFLAGS'.
In this case, those should be added into 'CFLAGS' or 'CXXFLAGS' to
configure Qt creator as same as sdk-terminal.

Signed-off-by: Junil Kim <jjunil79.kim@lge.com>

### DIFF
--- a/docs/articles/sdk/configure-qtcreator.rst
+++ b/docs/articles/sdk/configure-qtcreator.rst
@@ -124,7 +124,7 @@ Kit
 * Choose the previously defined PELUX C and C++ compilers from the corresponding combo boxes
 * Change the environment of this kit to the result of the ``env`` command ran in an SDK terminal
     * If ``CC`` and ``CXX`` have default compiler flags, those should be added in ``CFLAGS`` and ``CXXFLAGS``
-      To replace ``CFLAGS`` and ``CXXFLAGS``, use the result of the below command ran in
+      To replace ``CFLAGS`` and ``CXXFLAGS``, use the result of the below commands ran in
       an SDK terminal
 
       ``echo $CC | awk -v c="$CFLAGS" '{first = $1; $1 = ""; print "CFLAGS="$0, c;}'``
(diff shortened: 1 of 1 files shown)


sparkleholic:
This PR was created from this discussion. (https://github.com/Pelagicore/template-service/pull/9)

If you think there are awkward sentences or a better command to replace `CFLAGS/CXXFLAGS`, please leave a comment or modify it.

FlorentRevest (approved):

fhajredini (approved):

docs/articles/sdk/configure-qtcreator.rst
```diff
@@ -123,6 +123,12 @@ Kit
   the result of the ``echo ${SDKTARGETSYSROOT}`` command ran in an SDK terminal
 * Choose the previously defined PELUX C and C++ compilers from the corresponding combo boxes
 * Change the environment of this kit to the result of the ``env`` command ran in an SDK terminal
+    * If ``CC`` and ``CXX`` have default compiler flags, those should be added in ``CFLAGS`` and ``CXXFLAGS``
+      To replace ``CFLAGS`` and ``CXXFLAGS``, use the result of the below command ran in
```
thenor: s/command/commands/
s/ran/run/

thenor (changes requested):
Minor changes requested.
When I view it in the browser it looks like the commands end up on the same line, but I don't know if this is solved when we build the docs (as opposed to viewing it in Github).
Thanks for this!

docs/articles/sdk/configure-qtcreator.rst
sparkleholic: @thenor Thanks for the review. 
* `s/command/commands/` This seems to be fixed.
* `s/ran/run` <- Should I fix it? Because the above other similar sentences also use a same tense. If I'm wrong, please let me know. I'm ok to change this as well.

sparkleholic:
@thenor 
Unlike what we see here, it looks ok after the build.

thenor (approved):
I see that the tense is wrong for all these instances, but you don't need to change that, so just leave it as "ran". We will note down that we need to fix this.
Thanks!

thenor:
Note: this change will not be included in SWF until the git submodule is updated.

PR #132, opened 2018-07-20
Use apt instead of apt-get for consistency

sashko:
Signed-off-by: Oleksandr Kravchuk <oleksandr.kravchuk@pelagicore.com>


fhajredini:
Does apt work in older versions as well or is it for ubuntu 18.xx only?

sashko:
apt is a fairly old tool by now; we do not support anything below 16.04 anyway.

fhajredini (approved):

PR #133, opened 2018-07-24
Configure QtCreator: Adds a sentence about gdbserver usage

FlorentRevest:
This clarifies the remote debugging capabilities of Qt Creator.


docs/articles/sdk/configure-qtcreator.rst
```diff
@@ -157,6 +157,7 @@ How to open the template service project with QtCreator
 * Here you should select the kit to use to build project
 * Select the kit you have already created in the previous step and click ``Configure``
 * You can now compile and deploy to your target using the green arrow on the left of your screen
-
+* You can also remotely debug your program using the green arrow with a gray insect on the left of your screen.
```
sashko, line 160: grey

FlorentRevest: Actually, I checked and relied on http://www.dictionary.com/e/gray-or-grey/
Gray is more american and grey is more british... Not sure which one we should use for consistency, I don't think it has ever been defined

sashko: As a fellow European, I am on your side, but we used American English everywhere here :(

sashko (commented):
btw, title in present simple, please :-P

docs/articles/sdk/configure-qtcreator.rst
FlorentRevest: Then, we should stick with gray with an a ... ?

PR #134, opened 2018-09-07
Spellcheck: remove modals

thenor:
"Don't", "shouldn't", "didn't" etc. are caught in spellchecker and should be
avoided in the documentation.

Signed-off-by: Therese Nordqvist <therese.nordqvist@pelagicore.com>

### DIFF
--- a/docs/articles/sde/introduction_to_sde.rst
+++ b/docs/articles/sde/introduction_to_sde.rst
@@ -18,7 +18,7 @@ Dependencies:
 * VirtualBox
 * Self-extracting SDK package to install into the SDE. See, :ref:`building-the-sdk`.
 
-.. note:: Ubuntu 16.04 default repositories doesn't provide the latest version
+.. note:: Ubuntu 16.04 default repositories do not provide the latest version
    required. Please check the latest version of `Vagrant`_ and `VirtualBox`_ at those links.
 
 .. _`Vagrant`: https://www.vagrantup.com/downloads.html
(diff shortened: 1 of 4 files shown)


sashko (commented):
I've seen this yesterday. We should figure out why apostrophe isn't recognized instead.

docs/articles/sde/introduction_to_sde.rst
```diff
@@ -18,7 +18,7 @@ Dependencies:
 * VirtualBox
 * Self-extracting SDK package to install into the SDE. See, :ref:`building-the-sdk`.
 
-.. note:: Ubuntu 16.04 default repositories doesn't provide the latest version
+.. note:: Ubuntu 16.04 default repositories does not provide the latest version
```
TSAnsari: should be "do not"

sashko:
@thenor push the Jenkinsfile change and let's try to figure out what's wrong with the apostrophe on Monday.

docs/articles/sde/introduction_to_sde.rst
thenor: Yes. Missed this one. I will fix.

thenor:
I will close this and we can revisit the change later. I think it should be done, not because the spellchecker fails, but because modals shouldn't be used in documentation. On the other hand if we want to investigate why it fails then we can do that first.

martin-ejdestig:
> I've seen this yesterday. We should figure out why apostrophe isn't recognized instead.

I would argue the opposite. We should commit this change regardless of what we do with the spell checker. I do not think it is appropriate to use contractions in formal/technical writing.

>  I think it should be done, not because the spellchecker fails, but because modals shouldn't be used in documentation.

Yes. +1 for this.

sashko (approved):

PR #136, opened 2018-09-18
Fix taglist extension

demonicblue:
Make compatible with sphinx 1.8.0

### DIFF
--- a/docs/python_modules/sphinxexts/taglist.py
+++ b/docs/python_modules/sphinxexts/taglist.py
@@ -54,15 +54,6 @@ from docutils.parsers.rst import Directive
 from docutils import nodes
 from sphinx.environment import NoUri
 
-
-def initialize(app):
-    env = app.builder.env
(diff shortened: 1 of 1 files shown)


sashko:
Are we gonna upgrade sphinx to 1.8.x on our Docker image? Does it break compatibility with the currently used version?

demonicblue:
I would expect this extension to work with both 1.7.x and 1.8.x but the failed integration build says otherwise :(

thenor:
@demonicblue It seems to be a spelling error, so nothing serious.
"[Blueprint] Found 1 misspelled words"
We don't have a way to check where the error is from Jenkins, but if you build locally can you see the spelling error?

thenor (approved):
From what I can see the spelling error is from a previous merge (I'm looking at you @sashko :P).

sashko (approved):
Therese has tested it and confirmed it works on older versions as well; Jenkins build has failed due to a different issue. Feel free to merge it.

PR #140, opened 2018-10-22
Remove auto-update from example service files

DunderRoffe:
Remove the pre execute command stating docker pull in the example service
files for jenkins and docker-registry. This is due to it causing more
problems that benefit. For instance, sometimes one would like to restart
a service without also updating it.

Signed-off-by: Viktor Sjölind <vsjolind@luxoft.com>


thenor (approved):
I don't know if this is something we are using in our pelux CI environment, but I don't mind the change.

sashko:
> This is due to it causing more problems that benefit

What kind of?

Wouldn't it be a good practice to automatize security updates?

PR #141, opened 2018-10-23
Remove auto-update from example service files

DunderRoffe:
Remove the pre execute command stating docker pull in the example service
files for jenkins and docker-registry. This is due to it causing more
problems that benefit. For instance, sometimes one would like to restart
a service without also updating it.

Signed-off-by: Viktor Sjölind <vsjolind@luxoft.com>

### DIFF
--- a/docs/articles/infrastructure/ci-cd/docker-registry.rst
+++ b/docs/articles/infrastructure/ci-cd/docker-registry.rst
@@ -31,8 +31,6 @@ registry.
    TimeoutStartSec=0
    Restart=always
    ExecStartPre=-/usr/bin/docker stop %n
-   ExecStartPre=-/usr/bin/docker rm %n
-   ExecStartPre=/usr/bin/docker pull registry:2
    ExecStart=/usr/bin/docker run --rm --name %n \
              -e REGISTRY_HTTP_ADDR=0.0.0.0:80 \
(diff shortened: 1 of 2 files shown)


fhajredini (approved):
It seems that this page is not included in our software factory. It can only be found on the search bar, or if you go to the link directly. In case we need it, then maybe its a good idea to add it as well. 

Otherwise, I'm good with it.

DunderRoffe:
To be clear you do use jenkins.rst

thenor:
I don't feel that the removal of the "rm" line is described. I feel the commit message only concerns removal of the other line. I would appreciate an update to the commit message to clarify that part of the change a little more.

DunderRoffe:
@thenor rewritten

thenor (approved):

DunderRoffe:
Sorry I can't seem to access the error log. Can anyone say why this build fails?
@sashko @thenor

thenor:
@DunderRoffe The build view issue is a known problem with a Jenkins plugin. The build error is also a known error. It's a spelling error in another part (some word like "don't" instead of "do not"), so it has nothing to do with your changes.

thenor (approved):
Don't mind the build issue. It's not related to this change.

PR #143, opened 2018-10-29
Spellcheck: disable spellchecker

thenor:
Due to an issue we don't know how to solve we are disabling spellchecker.

Signed-off-by: Therese Nordqvist <therese.nordqvist@pelagicore.com>


fhajredini (approved):

thenor:
After some thought we won't solve this in this way. We want to have a spellchecker, so we will instead solve this in another way.

PR #148, opened 2018-11-12
dlt-logs: Add label

fhajredini:
Add label so that the dlt tutorial can be included on the other pages.


fhajredini:
Currently it fails because the changed file its not used anywhere and it throws the error that it isnt included in any toctree. However this file will be included in the Software Factory.

TSAnsari (approved):

sashko (approved):

PR #150, opened 2018-11-15
Move a file from swf to swf-blueprint

fhajredini:
Previously software factory blueprint was not building because of files
being included on swf instead of swf-blueprint. I have moved that file
from swf to software-factory-blueprint.

Also had to add `:orphan:` which allows for a file to not be included in
a toctree (which is neccesary).

Because I added that file to the swf-blueprint, spellcheck wordlist had
to be updated as well.


docs/spelling_wordlist.txt
```diff
@@ -347,3 +347,14 @@ initramfs
 efi
 partprobe
 lfs
+initramdisk
```
TSAnsari, line 350: Not related to this repository or commit. But this means some of these you would not need in software-factory's spelling word list. So make sure you remove them from software factory. Otherwise I am ok with it

TSAnsari (approved):

thenor (approved):

sashko (approved):
commit title though :(

docs/spelling_wordlist.txt
fhajredini: When building swf, it won't be searching for the words in swf-blueprint. Instead it looks for the words in swf(even for the pages that are under swf-blueprint submodule). Therefore all the words have to remain in swf as well.

PR #151, opened 2018-11-15
Spellcheck: add missing word

fhajredini:


fhajredini:
As weird as it is, this doesnt trigger any error when built locally. However it does on Jenkins job.

sashko (approved):

PR #153, opened 2018-12-04
Update FreeIPA instructions

demonicblue:
New instructions for setting up a replica.
Also a few clarifications.


docs/articles/infrastructure/ci-cd/freeipa.rst
```diff
@@ -171,6 +182,169 @@ fails.
...
+
+.. code-block:: none
+
+    kinit jdoe
+
+Then add the FQDN of the replica into the ipaservers group:
+
+.. code-block:: none
+
+    ipa hostgroup-add-member ipaservers --hosts replica1.example.test
```
frznlogic: the domain doesn't match the rest of the documentation.

docs/articles/infrastructure/ci-cd/freeipa.rst
```diff
@@ -171,6 +182,169 @@ fails.
...
+      [1/4]: stopping ntpd
+      [2/4]: writing configuration
+      [3/4]: configuring ntpd to start on boot
+      [4/4]: starting ntpd
+    Done configuring NTP daemon (ntpd).
+    Configuring directory server (dirsrv). Estimated time: 1 minute
+
+
+.. tip::
+    It's ok if ipa-replica-install gives error about "Reverse DNS resolution".
```
frznlogic: Even though I do understand this comment about "this error is ok, but this one is bad" it still chafes on me. 

Could we at least have a link or something to a "Troubleshooting guide" or something additionally at least? I'm not sure if I'd like to replace the entire tip with a link to it or just parts of it, or just an additional piece of information but... this is an very arbitrary list of errors right now.

docs/articles/infrastructure/ci-cd/freeipa.rst
```diff
@@ -171,6 +182,169 @@ fails.
...
+
+        [error] RuntimeError: failed to create ds instance Command '/usr/sbin/setup-ds --silent --logfile - -f /tmp/tmpuhBVEm' returned non-zero exit status 1
+
+    It may be that you either misconfigured /etc/hosts or need to add a reverse
+    DNS entry for the replica.
+
+Setup DNS on replica
+^^^^^^^^^^^^^^^^^^^^
+
+It's a good idea to setup DNS on at least on of your replicas and use it as a
```
frznlogic: "on at least one..."

docs/articles/infrastructure/ci-cd/freeipa.rst
```diff
@@ -316,4 +490,76 @@ Try to identify an LDAP user:
     # Should return "no such user"
     id someldapuser
 
+Replica
+^^^^^^^
+
+If any of the steps fails during installation you need to run 
+``ipa-server-install --uninstall`` a few times to clean the system. 2-3 should
+do it. After that you need to re-enroll the client before continuing with
```
frznlogic: This is awkward. Why does it need to be uninstalled 2-3 times?

docs/articles/infrastructure/ci-cd/freeipa.rst
```diff
@@ -316,4 +490,76 @@ Try to identify an LDAP user:
...
+    Update succeeded
+
+      [4/23]: creating installation admin user
+      [5/23]: setting up certificate server
+    ipa.ipaserver.install.cainstance.CAInstance: CRITICAL Failed to configure CA instance: Command '/usr/sbin/pkispawn -s CA -f /tmp/tmpkb9ORA' returned non-zero exit status 1
+    ipa.ipaserver.install.cainstance.CAInstance: CRITICAL See the installation logs and the following files/directories for more information:
+    ipa.ipaserver.install.cainstance.CAInstance: CRITICAL   /var/log/pki/pki-tomcat
+      [error] RuntimeError: CA configuration failed.
+
+It may mean that it cannot find java. You need to make sure /bin/java exists. If it doesn't, symlink it to the proper place. In Ubuntu 16.04:
```
frznlogic: This is being mentioned for the third time here I think. Could/should there be a troubleshooting section or something and all the problems be broken out into that one so that the "normal" workflow is handled first and if something breaks, it can be described in the troubleshooting section?

frznlogic (commented):
Minor nitpicks mainly really, but please look at them and say yay or nay

docs/articles/infrastructure/ci-cd/freeipa.rst
thenor:
```suggestion
    ipa hostgroup-add-member ipaservers --hosts replica1.example.com
```

docs/articles/infrastructure/ci-cd/freeipa.rst
```diff
@@ -171,6 +182,169 @@ fails.
...
+    Enter an IP address for a DNS forwarder, or press Enter to skip: 192.0.2.67
+    DNS forwarder 192.0.2.67 added. You may add another.
+    Enter an IP address for a DNS forwarder, or press Enter to skip: 
+    Checking DNS forwarders, please wait ...
+    Do you want to search for missing reverse zones? [yes]: 
+
+    The following operations may take some minutes to complete.
+    Please wait until the prompt is returned.
+
+If the installation fails, follow the instructions in the troubleshooting section
```
thenor: As @frznlogic has already mentioned, shouldn't all the above be part of the "Troubleshooting"? This is troubleshooting to me: issue description -> possible cause(s) and proposed solution(s)

docs/articles/infrastructure/ci-cd/freeipa.rst
thenor: I understand that since this is a how-to it's good to include this type of troubleshooting info, but I agree that something about the "it often fails and you need to uninstall many times to fix it" phrasing feels a bit off anyway. I don't have any good ideas on how to rephrase it though.

docs/articles/infrastructure/ci-cd/freeipa.rst
```diff
@@ -316,4 +490,76 @@ Try to identify an LDAP user:
...
+doesn't randomize the ID:s of the generated certificates. So the browser won't
+accept seeing a different certificate with the same Domain, Subject and Serial
+Number. To mitigate this you can either pass the an argument during installation
+to change the Subject or configure the Dogtag server to randomize ID:s.
+
+Set the subject:
+
+.. code-block:: bash
+
+    ipa-server-install --setup-dns --hostname=ipa.example.com.com --realm=IPA.EXAMPLE.COM --domain=ipa.example.com --subject="O=EXAMPLE.COM 20181112"
```
thenor: "hostname=ipa.example.com.com" is this extra "com" on purpose?

thenor (changes requested):
I think the biggest improvement would be to break-out and move the troubleshooting to the already existing chapter.

docs/articles/infrastructure/ci-cd/freeipa.rst
demonicblue: Yes, moving this to the troubleshooting section.

docs/articles/infrastructure/ci-cd/freeipa.rst
demonicblue: Good catch

docs/articles/infrastructure/ci-cd/freeipa.rst
demonicblue: Moving the error to the troubleshooting section. Also clarifying about the "false" error message.

docs/articles/infrastructure/ci-cd/freeipa.rst
demonicblue: Isn't that how it's currently handled? The java path gotcha is described in the installation step and then again in the troubleshooting section if you miss it.

I just wanted to include it for all relevant errors I came across. Because it was very unclear why the installation was failing in these cases and I wasted a lot of time finding out it was a missing path.

docs/articles/infrastructure/ci-cd/freeipa.rst
demonicblue: Please resolve if you're satisfied :)

docs/articles/infrastructure/ci-cd/freeipa.rst
demonicblue: Tried to give a reasonable explanation as to why. I can't change the underlying cause tho.

docs/articles/infrastructure/ci-cd/freeipa.rst
```diff
@@ -171,6 +182,163 @@ fails.
...
+goes down, clients will still be able to authenticate users against the replica.
+If the replica is also setup as a DNS and CA, the replica can be set as master
+in case the master is lost and disaster recovery has to be performed.
+
+Install required packages:
+.. code-block:: none
+
+    apt install -y freeipa-server freeipa-server-dns
+
+If you havn't already, enroll the replica server to the domain as described
```
thenor:
```suggestion
If you have not already, enroll the replica server to the domain as described
```

docs/articles/infrastructure/ci-cd/freeipa.rst
```diff
@@ -171,6 +182,163 @@ fails.
...
+
+    kinit jdoe
+
+Then add the FQDN of the replica into the ipaservers group:
+
+.. code-block:: none
+
+    ipa hostgroup-add-member ipaservers --hosts replica1.example.com
+
+Check that the result of command ipa hostgroup-find:
```
thenor:
```suggestion
Check that the result of command 'ipa hostgroup-find' is:
```

thenor: I remember there is some formatting we use for commands and paths etc. I think it is italic, but at the moment I can't remember if that is single ' or something else. If you write a command in-line this should be used, but it's not necessary inside code-blocks.

docs/articles/infrastructure/ci-cd/freeipa.rst
```diff
@@ -171,6 +182,163 @@ fails.
...
+    1 hostgroup matched
+    -------------------
+      Host-group: ipaservers
+      Description: IPA server hosts
+      Member hosts: ipa-server.example.com, replica1.example.com
+    ----------------------------
+    Number of entries returned 1
+    ----------------------------
+
+And also that ipa host-find returns both master and replica servers.
```
thenor: Same as the above comment here for the formatting of the command

docs/articles/infrastructure/ci-cd/freeipa.rst
```diff
@@ -171,6 +182,163 @@ fails.
...
+
+.. code-block:: none
+
+    ipa host-find
+
+On the replica server, modify /etc/hosts and add replica1:
```
thenor: ...and here for the path.

docs/articles/infrastructure/ci-cd/freeipa.rst
```diff
@@ -171,6 +182,163 @@ fails.
...
+
+.. tip::
+    It's ok if ipa-replica-install gives a warning message about "Reverse DNS
+    resolution". The replica will function without it.
+    
+
+Setup DNS on replica
+^^^^^^^^^^^^^^^^^^^^
+
+It's a good idea to setup DNS on at least one of your replicas and use it as a
```
thenor:
```suggestion
It is a good idea to setup DNS on at least one of your replicas and use it as a
```

docs/articles/infrastructure/ci-cd/freeipa.rst
```diff
@@ -171,6 +182,163 @@ fails.
...
+
+On the replica server, modify /etc/hosts and add replica1:
+
+.. code-block:: none
+    :caption: Example /etc/hosts
+
+    192.0.2.6 replica1.example.com replica1
+
+If you use FreeIPA as your DNS, you may also want to add a reverse DNS entry for
+the replica but it's not necessary.
```
thenor:
```suggestion
the replica but it is not necessary.
```

docs/articles/infrastructure/ci-cd/freeipa.rst
thenor: Could you add a link to the Replica Troubleshooting section from here?

docs/articles/infrastructure/ci-cd/freeipa.rst
```diff
@@ -171,6 +182,163 @@ fails.
...
+.. code-block:: none
+
+    dig ipa-server.example.com @127.0.0.1
+    dig ipa-server.example.com @192.0.2.5
+
+Setup CA on replica
+^^^^^^^^^^^^^^^^^^^
+
+Make sure /bin/java exists, otherwise symlink it to the proper binary. On ubuntu
+16.04 it's /usr/bin/java . Otherwise it will cause the certificate server to not
```
thenor:
```suggestion
16.04 it is /usr/bin/java . Otherwise it will cause the certificate server to not
```
Also format the path to match formatting rules.

docs/articles/infrastructure/ci-cd/freeipa.rst
```diff
@@ -171,6 +182,163 @@ fails.
...
+
+    ln -s /usr/bin/java /bin/java
+
+Then have the Directory Manager password ready and start the installation:
+
+.. code-block:: none
+
+    ipa-ca-install
+
+If the installation fails, follow the instructions in the troubleshooting section
```
thenor: A link here to the Troubleshooting section which is meant.

docs/articles/infrastructure/ci-cd/freeipa.rst
```diff
@@ -316,4 +484,87 @@ Try to identify an LDAP user:
     # Should return "no such user"
     id someldapuser
 
+Replica
+^^^^^^^
+
+If any of the steps fails during installation you need to run
```
thenor:
```suggestion
If any of the steps fail during installation you need to run 
```

docs/articles/infrastructure/ci-cd/freeipa.rst
```diff
@@ -316,4 +484,87 @@ Try to identify an LDAP user:
...
     id someldapuser
 
+Replica
+^^^^^^^
+
+If any of the steps fails during installation you need to run 
+``ipa-server-install --uninstall`` a few times to clean the system. 2-3 should
+do it. This is to ensure all configurations are cleared if you tried installing
+multiple times. After that you need to re-enroll the client before continuing
+with ``ipa-client-install --enable-dns-updates --force-join`` (Don't forget to
```
thenor:
```suggestion
with ``ipa-client-install --enable-dns-updates --force-join`` (Do not forget to
```

docs/articles/infrastructure/ci-cd/freeipa.rst
```diff
@@ -316,4 +484,87 @@ Try to identify an LDAP user:
...
+re-enable mkhomedir).
+
+If you find yourself re-installing FreeIPA server you will run into certificate
+problems when accessing the Web UI. This is because the certificate server
+doesn't randomize the ID:s of the generated certificates. So the browser won't
```
thenor:
```suggestion
does not randomize the ID:s of the generated certificates, this is since the browser will not
```

docs/articles/infrastructure/ci-cd/freeipa.rst
```diff
@@ -316,4 +484,87 @@ Try to identify an LDAP user:
...
+accept seeing a different certificate with the same Domain, Subject and Serial
+Number. To mitigate this you can either pass the an argument during installation
```
thenor:
```suggestion
Number. To mitigate this you can either pass an argument during installation
```

docs/articles/infrastructure/ci-cd/freeipa.rst
```diff
@@ -316,4 +484,87 @@ Try to identify an LDAP user:
...
+``ipa-server-install --uninstall`` a few times to clean the system. 2-3 should
+do it. This is to ensure all configurations are cleared if you tried installing
+multiple times. After that you need to re-enroll the client before continuing
+with ``ipa-client-install --enable-dns-updates --force-join`` (Don't forget to
+re-enable mkhomedir).
+
+If you find yourself re-installing FreeIPA server you will run into certificate
+problems when accessing the Web UI. This is because the certificate server
+doesn't randomize the ID:s of the generated certificates. So the browser won't
+accept seeing a different certificate with the same Domain, Subject and Serial
```
thenor:
```suggestion
accept different certificates with the same Domain, Subject and Serial
```
Maybe? I'm assuming this means that if it finds a second certificate with the same Domain/Subject/Serial number it won't like it.

docs/articles/infrastructure/ci-cd/freeipa.rst
```diff
@@ -316,4 +484,87 @@ Try to identify an LDAP user:
     # Should return "no such user"
     id someldapuser
 
+Replica
+^^^^^^^
```
thenor: I would prefer some headers within this section. If possible, it makes it easier to scan through faster without having to text-search.

docs/articles/infrastructure/ci-cd/freeipa.rst
```diff
@@ -316,4 +484,87 @@ Try to identify an LDAP user:
...
+    Update succeeded
+
+      [4/23]: creating installation admin user
+      [5/23]: setting up certificate server
+    ipa.ipaserver.install.cainstance.CAInstance: CRITICAL Failed to configure CA instance: Command '/usr/sbin/pkispawn -s CA -f /tmp/tmpkb9ORA' returned non-zero exit status 1
+    ipa.ipaserver.install.cainstance.CAInstance: CRITICAL See the installation logs and the following files/directories for more information:
+    ipa.ipaserver.install.cainstance.CAInstance: CRITICAL   /var/log/pki/pki-tomcat
+      [error] RuntimeError: CA configuration failed.
+
+It may mean that it cannot find java. You need to make sure /bin/java exists. If it doesn't, symlink it to the proper place. In Ubuntu 16.04:
```
thenor:
```suggestion
It may mean that it cannot find java. You need to make sure /bin/java exists. If it does not, symlink it to the proper place. In Ubuntu 16.04:
```

thenor (changes requested):
Some nit-picky comments. Mostly on removal of "doesn't", "won't" etc.

docs/articles/infrastructure/ci-cd/freeipa.rst
demonicblue: Good catch

docs/articles/infrastructure/ci-cd/freeipa.rst
demonicblue: Added that but it starts to feel like an american research paper with infinite levels of subsections.

thenor: Haha sorry. I don't feel it's necessary, but I prefer it since it usually makes the process of troubleshooting a little easier.

thenor (approved):
Looks good to me!

PR #163, opened 2019-05-31
booting-qemu-image: Add runqemu details and build instructions

fhajredini:


JEderonn (approved):

docs/articles/baseplatform/booting-qemu-image.rst
```diff
@@ -2,22 +2,81 @@
...
+------------------
+
+``qemu_launcher`` aims for the same goal as ``runqemu`` but offers more flexibility. 
 
 Images for QEMU target can be booted up using KVM [#kvm]_ to utilize hardware
 virtualization feature if found on the host system and run the virtual
 machine in better performance.
 See KVM feature for QEMU [#qemu_kvm_feature]_.
 
-QEMU launch script
```
sashko, line 13: why did you remove the header?

docs/articles/baseplatform/booting-qemu-image.rst
```diff
@@ -2,22 +2,81 @@
...
+.. code-block:: bash
+
+          $  runqemu
+
+Now on another terminal of the same machine, you should be able to ssh ``root@192.168.7.2``. 
+
+QEMU launch script
+------------------
+
+``qemu_launcher`` aims for the same goal as ``runqemu`` but offers more flexibility.
```
JEderonn: like what flexibility, just state some examples

docs/articles/baseplatform/booting-qemu-image.rst
fhajredini: its on line 63 now..

fhajredini: or do you mean something else?

fhajredini:
I will merge this by the end of the day unless there are further comments!

docs/articles/baseplatform/booting-qemu-image.rst
```diff
@@ -2,22 +2,81 @@
...
+####################################
+Building and booting up a QEMU image
+####################################
+
+Booting a QEMU image can be using one of the two scripts
+
+- ``runqemu``
+- ``qemu_launcher``
+
+Both are scripts but they are offered from different layers. ``runqemu`` is a part of OpenEmbedded while qemu_launcher is a part of meta-pelux. ``qemu_launcher`` requires enabling the virtualization on BIOS while ``runqemu`` is more straight forward and simple. Both options are described below.
```
mardy: "a part" -> "part" (in two occurrences)

"straight forward" -> "straightforward"

Also, please break long lines at 80 chars.

docs/articles/baseplatform/booting-qemu-image.rst
```diff
@@ -2,22 +2,81 @@
...
+
+Run qemu script
+---------------
+Finally, you can ``runqemu`` 
+
+.. code-block:: bash
+
+          $  runqemu
+
+Now on another terminal of the same machine, you should be able to ssh ``root@192.168.7.2``.
```
mardy: to ssh -> to run ``ssh`` (move the ssh inside the monospace font)

docs/articles/baseplatform/booting-qemu-image.rst
```diff
@@ -2,22 +2,81 @@
...
+.. code-block:: bash
+
+          $  runqemu
+
+Now on another terminal of the same machine, you should be able to ssh ``root@192.168.7.2``. 
+
+QEMU launch script
+------------------
+
+``qemu_launcher`` aims for the same goal as ``runqemu`` but is aimed more towards KVM. It wraps up the kvm command and provides simple options to launch dev and prod images without loosing the flexibility to adjust parameters like memory, cpu, etc.
```
mardy: no need to abbreviate dev and prod

mardy (changes requested):
Just some nitpicking :-)

docs/articles/baseplatform/booting-qemu-image.rst
fhajredini: I did that change, but how about removing the development and production entirely, since for QEMU we only provide the minimal and minimal-dev images. Also neither one of them is production..

mardy: Good point! :-)

martin-ejdestig: So, I do not really understand why qemu_launcher is needed. As far as I know QEMU uses KVM if it is available. (If the `runqemu` script does not use KVM by default if it is available, one just has to append `kvm` to the command line, it looks like.)

And `qemu_launcher` does not use QEMU? I don't know... the more I look at it the more confused I get. Commit message nor #97 contains any information as to why it is needed and why it should be used instead of `runqemu`. :(

(Asking this question here since I am not sure adding a section for it makes sense... if it is not needed. :)

martin-ejdestig: If the `runqemu` script does not use KVM by default if it is available, one just has to append `kvm` to the command line, it looks like.

martin-ejdestig: And `qemu_launcher` does not use QEMU? I don't know... the more I look at it the more confused I get.

martin-ejdestig: Commit message nor #97 contains any information as to why it is needed and why it should be used instead of `runqemu`. :(

martin-ejdestig: Anyway, not blocking this to get merged. I am just confused and wonder if it really makes sense to have it at all... and if we can't explain to ourselves why it is needed, how can we do so to someone reading this section. :)

fhajredini: >As far as I know QEMU uses KVM if it is available.

That happens when u start QEMU with `runqemu`. On the other hand, if you want to use qemu_launcher, you must have KVM. 

>  qemu_launcher does not use QEMU?

It does.. I dont get it at what point did this confuse you?

> Commit message nor #97 contains any information 

Are you sure u meant about #97? It doesnt seem relevant to this PR. 
However, my understanding of why we should use qemu_launcher is that it has been developed internally.

fhajredini: Generally speaking tho, i was tempted to remove any information about qemu_launcher, but after some discussion it was decided to only add the information about runqemu and leave the topic of qemu_launcher for advanced users.

martin-ejdestig: > > As far as I know QEMU uses KVM if it is available.
> 
> That happens when u start QEMU with `runqemu`. On the other hand, if you want to use qemu_launcher, you must have KVM.
> 

Yes, so what is the point?

> > qemu_launcher does not use QEMU?
> 
> It does.. I dont get it at what point did this confuse you?

Nope, it does `sudo kvm ...` . "QEMU" is only used in help strings etc.

> > Commit message nor #97 contains any information
> 
> Are you sure u meant about #97? It doesnt seem relevant to this PR.

Sorry, the PR is in meta-pelux: https://github.com/Pelagicore/meta-pelux/pull/97 .

fhajredini: So you recommending to remove the section about qemu_launcher entirely? I dont really have a strong opinion on that and i would agree that it only confuses the user.

fhajredini:
For now, im leaving the qemu_launcher instructions as no final decision has been taken.

PR #164, opened 2019-06-24
deploying-image.rst: Include bmap instructions

fhajredini:
Signed-off-by: Fisnik Hajredini <fhajredini@luxoft.com>

PR #167, opened 2019-08-01
deploying the image: add information about bmaptool usage

pancio88:


martin-ejdestig:
Looks like mention of bmap tools were done in #164 .

@pancio88 Could you perhaps update to latest master and see if you think the instructions need to be made more clear? E.g. something that was not added in #164 but you have in your commit is a link to https://source.tizen.org .

pancio88:
> Looks like mention of bmap tools were done in #164 .
> 
> @pancio88 Could you perhaps update to latest master and see if you think the instructions need to be made more clear? E.g. something that was not added in #164 but you have in your commit is a link to https://source.tizen.org .

Sure, I realize now that the version used into my project is not updated. Thanks for pointing out.

pancio88:
@martin-ejdestig I also changed a bit about the rootfs mention/compression alg. wic image can be compressed by yocto using also gz and is not necessary to point out rootfs since can contains also other partition edit from the .wks file.

fhajredini (approved):

